PR #21, opened 2015-03-19
Password verification via login.webmaker.org api

cadecairos:


cadecairos:
How does this look, @jbuck / @alicoding

jbuck:
@cadecairos r+

PR #24, opened 2015-03-20
Add form design v1

### LINKED ISSUE #6: Implement sign-in page
- [x] page stub
- [x] client-side username/password validation/acceptance
  - [x] usernames: at least 1 character
  - [x] password at least 8 characters
- [x] server-side username/password acceptance (?)
- [x] expose a URI for this page

alicoding:
fixes #6


templates/components/form.jsx
```diff
@@ -0,0 +1,98 @@
...
+    firstName: Joi.string().required().label('First Name'),
+    lastName: Joi.string().allow(null).label('Last Name'),
+    email: Joi.string().email().label('Email Address'),
+    username:  Joi.string().alphanum().min(3).max(30).required().label('Username'),
+    password: Joi.string().regex(/[a-zA-Z0-9]{3,30}/).label('Password'),
+    verifyPassword: Joi.any().valid(Joi.ref('password')).required().label('Password Confirmation')
+  },
+  getInitialState: function() {
+    return {
+      firstName: null,
```
Pomax: probably better to make these empty strings?

templates/components/form.jsx
```diff
@@ -0,0 +1,98 @@
...
+      lastName: null,
+      email: null,
+      username: null,
+      password: null,
+      verifyPassword: null,
+      feedback: null
```
Pomax: we have a feedback entry, but no validation for it.

templates/components/form.jsx
```diff
@@ -0,0 +1,98 @@
...
+    };
+  },
+  render: function() {
+     var that = this;
+     var fields = Object.keys(this.props.fields).map(function(key,i) {
```
Pomax: we can pull the function we pass into `map` as a component function instead (e.g. `mapfunction`) and then call it as `...map(this.mapfunction)`, and then because React does `this` rebinding, we don't need the self=this variable. It can stay "this" inside the mapping function

templates/components/form.jsx
```diff
@@ -0,0 +1,98 @@
...
+      <span className="help-block">Invalid Username or Password</span>
+    );
+  },
+  getClasses: function(field) {
+    return React.addons.classSet({
+      'form-control': true,
+      'has-error': !this.isValid(field),
+    });
+  },
+  getIconClass: function(field) {
```
Pomax: Based on https://facebook.github.io/react/docs/component-specs.html#statics  we should be able to use this here:
```
  getIconClass: function(field) {
    return Signup.phoneLabels[field];
  },
```

PR #44, opened 2015-03-24
login and routes

Pomax:
adds login as page, and stubs for sign up, password reset, an a general 404, with react-router routing tied back into Hapi


templates/lib/routes.jsx
```diff
@@ -0,0 +1,24 @@
+var React = require('react');
+var Router = require('react-router');
+var Route = Router.Route;
+var Link = Router.Link;
+var DefaultRoute = Router.DefaultRoute;
```
alicoding: Lets remove this?

alicoding:
once this is green lets land this ! r+ from me

PR #50, opened 2015-03-24
Fix issue46 - only ignore specific files

### LINKED ISSUE #46: Instead of force committing to public/, ignore specific files
Personally I think we should optimize for newly added static files, such as images, instead of optimizing for ignored files (which won't change often)

You could simplify the `gitignore` entry by adding everything that needs to be ignored to a specific folder *inside* public, such as `compiled/`, but I don't really think that's necessary

alicoding:
Closes #46

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 node_modules/
-public/
+app.bundle.*
+styles.css*


alicoding:
@k88hudson R?

k88hudson:
Looks good to me

alicoding:
Thanks!

PR #52, opened 2015-03-24
Added authorization codes

jbuck:


lib/oauth-db.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  process.nextTick(function() {
+    callback(null, this.clients[client_id]);
+  }.bind(this));
+};
+
+OAuthDB.prototype.generate_auth_code = function(client_id, user_id, scopes, expires_at, callback) {
+  crypto.randomBytes(32, function(err, random) {
+    Hoek.assert(!err, err);
```
thisandagain: Not super familiar with `hoek.assert` yet but this will throw an exception, yes? If so, I'm curious why we would throw here rather than passing it to the callback ala

```js
if (err) return callback(err);
```

thisandagain: Certainly an edge case (I hope we won't drain the entropy sources by generating 32 bytes!) but I suppose I'm curious about how we pattern when to throw vs. pass to the callback in our API design.

cadecairos: I believe we'll change that once the test script doesn't fail on <100% test coverage.

We couldn't think of a way to easily fake randomBytes failing, just to get test coverage in an `if` block.

cadecairos:
R+ from me

lib/oauth-db.js
thisandagain: Got it. That makes sense. Thanks for clarifying!

PR #59, opened 2015-03-24
Adding mofo-style and fixing bustifications

simonwex:
# In progress

_Outstanding_
* [x] Fix style issues
* [x] fix outstanding bugs


simonwex:
I think this last one takes care of tests. 

`/me waits for Travis`

lib/account.js.b
```diff
@@ -0,0 +1,59 @@
+var hyperquest = require('hyperquest');
```
alicoding: hmm what is this `.js.b` ?

simonwex: Ah, an artifact that shouldn't be there. Good catch.

cadecairos:
R+

PR #60, opened 2015-03-24
[WIP] signup page initial pass

Pomax:
do not merge yet


templates/less/pages/signup.less
```diff
@@ -0,0 +1,111 @@
+.signup-page {
+
+  .navbar {
+    background: transparent;
+  }
+
+  @white: /* not #F9F7FF...? */ white;
```
alicoding: Should we put this in `variables.less`?

templates/less/pages/signup.less
```diff
@@ -0,0 +1,111 @@
...
+      }
+      span {
+        font-weight: normal;
+      }
+    }
+
+    #feedback {
+      display: none;
+
+      & + span {
```
alicoding: `& +` ?

templates/components/form.jsx
```diff
@@ -1,44 +1,58 @@
...
   mixins: [
     ValidationMixin,
     React.addons.LinkedStateMixin
   ],
-  validatorTypes:  {
-    username: Joi.string().min(1).required().label('Username'),
-    password: Joi.string().min(3).required().label('Password')
+  validatorTypes: false,
+  componentWillMount: function() {
+    this.validatorTypes = this.props.validators;
```
alicoding: Should this ever be remove on `componentWillUnMount`? or whatever other page uses this will override and that would be enough?

templates/less/pages/signup.less
Pomax: we should, but as a followup: there are a LOT of values in our less files that should be variables but aren't.

templates/less/pages/signup.less
Pomax: `&` is the LESS code for "the current selector", so here it unpacks to `#feedback + span`, i.e. a span that follows the `<input id=feedback` element

templates/components/form.jsx
Pomax: remove what? `this` gets cleaned up after unmount, so there shouldn't be any lingering data. If these were event handlers: yes. As just plain properties, there should be no reason to.

templates/components/header.jsx
```diff
@@ -4,8 +4,13 @@ var Header = React.createClass({
...
       <div className="navbar">
-        <img src="/assets/img/logo.png" alt="logo" className="logo" />
-        <div className="sign-up">Need an account? <a href="" className="underline">Sign up</a></div>
+        {
+          !!this.props.wordmark ?
+            <img src="/assets/img/webmaker-horizontal.svg" alt="Mozilla Webmaker" className="wordmark" />
+            :
+            <img src="/assets/img/logo.png" alt="Mozilla Webmaker" className="logo" />
+        }
+        <div className="redirect">{this.props.redirectText} <a href={this.props.redirectURL} className="underline">{this.props.redirectLabel}</a></div>
```
alicoding: Can you update href to use this instead? 

```
<Link to={this.props.redirectPage} className="underline">{this.props.redirectLabel}</Link>
```

instead of 

```
<a href={this.props.redirectURL} className="underline">{this.props.redirectLabel}</a>
```

This way it will not do an actual page load redirect.

templates/pages/login.jsx
```diff
@@ -3,32 +3,38 @@ var React = require('react');
...
-        'password': {
-          'placeholder': 'Password',
-          'type': 'password'
-        }
-      }
-    ];
     return (
       <div>
-        <Header />
+        <Header redirectText="Need an account?" redirectLabel="Sign up" redirectURL="/signup" />
```
alicoding: Change
+        <Header redirectText="Need an account?" redirectLabel="Sign up" redirectURL="/signup" />

to 

+        <Header redirectText="Need an account?" redirectLabel="Sign up" redirectPage="signup" />

templates/pages/signup.jsx
```diff
@@ -1,11 +1,68 @@
...
+var validators = require('../lib/validatorset');
+var fieldValidators = validators.getValidatorSet(fieldValues);
 
 var Signup = React.createClass({
+
   render: function() {
     return (
-      <div>Signup page</div>
+      <div className="signup-page">
+        <Header wordmark redirectText="Already have an account?" redirectLabel="Log in" redirectURL="/login" />
```
alicoding:
```diff
- <Header wordmark redirectText="Already have an account?" redirectLabel="Log in" redirectURL="/login" />
+ <Header wordmark redirectText="Already have an account?" redirectLabel="Log in" redirectPage="login" />

alicoding:
Just 3 things to fix then r+

templates/components/header.jsx
Pomax: actually, cassie change the comps. there, this should *always* be the logo + wordmark now

PR #64, opened 2015-03-24
Adding tasks to README

gvn:


README.md
```diff
@@ -13,6 +13,19 @@ npm install
 npm start
 ```
 
+## Tasks
+
+- `npm start` Run web server.
+- `npm test` Run automated tests.
+- `npm build` Build project.
```
alicoding: This should be `npm run build` and I'm wondering if we should change the task to run both `less` and `js` because right now it only compile `js` here.

README.md
```diff
@@ -13,6 +13,19 @@ npm install
...
+- `npm run watch-css` Build LESS files when they change.
+- `npm run watch-js` Build JS files when they change.
+- `npm run watch` Build JS and LESS files when they change.
```
alicoding: You might need to rebase now since we don't have `watch` anymore but we are using `start` for this now.

README.md
```diff
@@ -13,6 +13,19 @@ npm install
 npm start
 ```
 
+## Tasks
+
+- `npm start` Run web server.
```
alicoding: Same as the `watch` one we are now running three tasks for this task which is running the web server, compile LESS and JS then start watch

alicoding:
Just need to rebase to match our current master branch with small tweaks and this will be good to land.

README.md
gvn: Yeah I think it should build everything. :+1:

README.md
gvn: What about?

`watch`
`watch:css`
`watch:js`

gvn: `npm run start` seems awkward to me...

README.md
gvn: What about calling it `dev` instead since that's what we called that task in Grunt/Gulp?

README.md
alicoding: I like the name suggestion, but also when we run we actually do `npm start`.

README.md
alicoding: We probably need the `npm start` anyway since we have to use this for heroku (as default start task by node instance) We make `npm run dev` for running web server and start watching, but `npm start` only **build** and start **web server**?

gvn:
OMG I have done something terrible to this branch...

PR #71, opened 2015-03-25
access_token route

cadecairos:


cadecairos:
R? @jbuck

web/server.js
```diff
@@ -102,8 +102,73 @@ module.exports = function(options) {
...
+                }
+                if ( client.client_secret !== request.payload.client_secret ) {
+                  return reply(Boom.forbidden('Invalid Client Credentials'));
+                }
+                reply(client);
+              });
+            }
+          },
+          {
+            assign: 'redirectUri',
```
jbuck: I was actually speccing out that we ignore any application-submitted redirect url entirely, and only use the registered one

web/server.js
```diff
@@ -102,8 +102,73 @@ module.exports = function(options) {
...
+                Date.now() + (1000 * 60 * 60 * 24),
+                reply
+              );
+            }
+          }
+        ]
+      },
       handler: function(request, reply) {
-        reply('ok');
+        var redirectObj = url.parse(request.pre.redirectUri, true);
```
jbuck: so instead you could use `request.pre.client.redirect_uri`

jbuck:
you can remove the variable redirect_uri bit. That'll simplify things a bit!

jbuck:
@cadecairos r+

PR #72, opened 2015-03-25
Add first pass at OAuth2 documentation

jbuck:


docs/oauth.md
```diff
@@ -0,0 +1,169 @@
...
+
+Webmaker requires the following information to register your application:
+
+* **Name** - The friendly name of your application e.g. "Buckley's Bees".
+* **Description** - A description of your application e.g. "Buckley's Bees sells the best honey in Ontario".
+* **Image** - An image that represents your application.
+* **Homepage URL** - The full URL to your application homepage e.g. "https://buckleysbees.ca".
+* **Redirect URL** - Your application's OAuth2 callback URL e.g. "https://buckleysbees.ca/oauth2/callback".
+
+Once you've registered your application you'll recieve your application credentials:
```
Pomax: This is still missing a bit on *how* or *where* to perform this registration (even if we don't have that implemented yet, let's note where we *intend* to put it and how that would work)

Pomax: secondary question: what format are the returned values in?

docs/oauth.md
```diff
@@ -0,0 +1,169 @@
...
+
+* **Client ID** - A string identifying your application. This value is not secret and can be shown publicly e.g. `wm_id_BTQNPABtUqqApaDrcsDa`.
+* **Client Secret** - A string. This value is secret and must not be shared with anyone e.g. `wm_secret_iHmAWhvCwBYnuE6aZHcArBEPohanKmmWr8LcyBBPZYRkAxdmui`.
+
+You can register your application by emailing [jbuck](mailto:jon@mozillafoundation.org). Self-serve application registration is coming shortly.
```
Pomax: okay so yes, that's the current way to do it, but if we have any idea how to do it later, let's already doc that =)

docs/oauth.md
```diff
@@ -0,0 +1,169 @@
...
+## Web Application Flow
+
+The web application flow is ideal for services that have a server-side component to securely store the client secret and user's access tokens.
+
+### 1. Redirect users to request access
+
+To start the web application flow you need to redirect your users browser to Webmaker.
+
+`GET https://id.webmaker.org/login/oauth/authorize`
+
```
Pomax: probably good to be a little more explicit here and say that as oauth2 description this is "GET" but in most cases this will be the website actually redirecting the user to another website, not an AJAX "GET" request or the like.

docs/oauth.md
```diff
@@ -0,0 +1,169 @@
...
+  	<td>client_id</td>
+  	<td>The client ID you received when you registered your application.</td>
+  </tr>
+  <tr>
+  	<td>response_type</td>
+  	<td>The response_type must be set to "code".</td>
+  </tr>
+  <tr>
+  	<td>scopes</td>
+  	<td>A space separated list of scopes.</td>
```
Pomax: which values can we choose from?

docs/oauth.md
```diff
@@ -0,0 +1,169 @@
...
+  </tr>
+  <tr>
+  	<td>state</td>
+  	<td>An unguessable random string. It is used to protect against cross-site request forgery attacks.</td>
```
Pomax: s/unguessable/secure/ ? Also, probably good to mention this should be random every time, because it is used in a handshake round trip, with the value going into the request, the oauth2 doing things and sending it back during callback, so that the requester can verify that the answer it got comes with the same "state" that it sent out?

docs/oauth.md
```diff
@@ -0,0 +1,169 @@
...
+  	<td>client_id</td>
+  	<td>The client ID you received when you registered your application.</td>
+  </tr>
+  <tr>
+  	<td>code</td>
+  	<td>A temporary authorization code</td>
+  </tr>
+  <tr>
+  	<td>state</td>
+  	<td>An unguessable random string. It is used to protect against cross-site request forgery attacks.</td>
```
Pomax: This is no longer "an", this is now "the random state value that you signed the original request with. If this value does not match the value sent with your initial request, callback processing should be cut off." or something to that effect

docs/oauth.md
```diff
@@ -0,0 +1,169 @@
...
+  	<th>Name</th>
+  	<th>Description</th>
+  </tr>
+  <tr>
+  	<td>client_id</td>
+  	<td>The client ID you received when you registered your application.</td>
+  </tr>
+  <tr>
+  	<td>response_type</td>
+  	<td>The response_type must be set to "code".</td>
```
cadecairos: what else could this be other than 'code'

cadecairos: ...assuming this is included because it can be expanded on in the future

PR #79, opened 2015-03-26
Wire up the loginz

cadecairos:


cadecairos:
r?

templates/pages/login.jsx
```diff
@@ -47,6 +49,17 @@ var Login = React.createClass({
   },
   handleFormData: function(error, data) {
     console.log("inside App we see:", error, data);
+    fetch('/login', {
+      method: 'post',
+      headers: {
+        'Accept': 'application/json',
```
jbuck: We should probably make this charset utf-8

templates/pages/login.jsx
```diff
@@ -47,6 +49,17 @@ var Login = React.createClass({
...
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        uid: data.username,
+        password: data.password
+      })
+    });
```
jbuck: Also, can you make this redirect to /login/oauth/authorize with the parameters we specified before?

jbuck:
This is looking good but needs a little bit more additional work!

web/index.js
```diff
@@ -4,7 +4,9 @@ var options = {
   host: process.env.HOST,
   port: process.env.PORT,
   loginAPI: process.env.LOGINAPI,
-  oauth_clients: process.env.OAUTH_DB ? JSON.parse(process.env.OAUTH_DB) : []
+  oauth_clients: process.env.OAUTH_DB ? JSON.parse(process.env.OAUTH_DB) : [],
+  cookieSecret: process.env.COOKIE_SECRET,
+  secureCookies: process.env.SECURE_COOKIES === true
```
jbuck: should be `=== "true"` since everything that comes through on `process.env` is a string

jbuck:
Can you add the new config values to the README as well?

templates/pages/login.jsx
```diff
@@ -46,7 +49,30 @@ var Login = React.createClass({
...
+      },
+      body: JSON.stringify({
+        uid: data.username,
+        password: data.password
+      })
+    }).then(function(response) {
+      var redirectObj;
+      if ( response.status === 200 ) {
+        redirectObj = Url.parse('/login/oauth/authorize', true);
+        redirectObj.query = queryObj;
```
jbuck: I think you also need to set search to null right? From the [docs](https://nodejs.org/api/url.html#url_url_format_urlobj):

```
search will be used in place of query.

    It is treated the same with or without the leading ? (question mark)

query (object; see querystring) will only be used if search is absent.
```

cadecairos: there shouldn't ever be a search, since the parse method is parsing just the absolute url to authorize

jbuck:
r+

PR #80, opened 2015-03-26
Reset password

alicoding:
#7


alicoding:
r?

templates/less/components/form.less
```diff
@@ -27,43 +28,58 @@
...
+  .icon-label-valid:after {
+    .label(@fa-check);
+    right: 10px;
+    top: 17px;
+    color: #53C0A3;
   }
 }
 .has-error {
-  border: 1px solid #E95B95;
+  border: 1px solid #E95B95 !important;
```
alicoding: yep I needed this to make sure it is red and not green when there is an error state >.<

templates/components/form/form.jsx
```diff
@@ -25,30 +27,47 @@ var Form = React.createClass({
...
+      email: null,
+      checked: false,
+      dirty: {}
     };
   },
   buildFormElement: function(key, i) {
-    var id = Object.keys(this.props.fields[i]);
+    // we always expect this.props.fields[i] to be one object with one property.
+    var self = this;
+    var id = Object.keys(this.props.fields[i])[0];
```
Pomax: extract as a `var values = var placeholder = this.props.fields[i][id];` and then just use straight up `values.type`, `values.label` etc? This is way too much this.props access it feels like

templates/components/form/form.jsx
```diff
@@ -25,30 +27,47 @@ var Form = React.createClass({
...
+    var isDisabled = this.props.fields[i][id].disabled;
+    var focus = this.props.fields[i][id].focus;
+    var username = this.props.fields[i][id].username;
+    this.passChecked = this.props.fields[i][id].checked;
 
-    function foo() {
-      console.log('foo');
-      this.handleValidation(id);
+    function dirty(err, valid) {
+      var state = self.state.dirty;
```
Pomax: you're not extracting "state", but the thing that records which elements are dirty. `var dirty` ?

templates/components/form/form.jsx
```diff
@@ -25,30 +27,47 @@ var Form = React.createClass({
...
+    var errorMessage = this.props.fields[i][id].errorMessage;
+    var isDisabled = this.props.fields[i][id].disabled;
+    var focus = this.props.fields[i][id].focus;
+    var username = this.props.fields[i][id].username;
+    this.passChecked = this.props.fields[i][id].checked;
 
-    function foo() {
-      console.log('foo');
-      this.handleValidation(id);
+    function dirty(err, valid) {
```
Pomax: this should not be a function that is redeclared every time `render` is called, it needs to be moved out (and then we can ditch `self`, too, it'll be autobound to the component)

templates/components/form/form.jsx
```diff
@@ -25,30 +27,47 @@ var Form = React.createClass({
...
     var input = (
       <input type={type}
              id={id}
+             ref={id+'Input'}
              placeholder={placeholder}
              valueLink={this.linkState(id)}
-             onBlur={this.handleValidation(id)}
-             className={this.getClasses(id)}
-             ref={id} />
+             onBlur={this.handleValidation(id, dirty)}
```
Pomax: `dirty` would become `this.dirty`

templates/components/form/form.jsx
```diff
@@ -25,30 +27,47 @@ var Form = React.createClass({
...
+             className={this.getInputClasses(id)}
+             defaultValue={username}
+             disabled={isDisabled ? "disabled" : false}
+             autoFocus={focus ? true : false}
+      />
+    );
+
+    var errorTooltip = (
```
Pomax: move down a few lines to after the type check, so we can keep those grouped?

templates/components/form/form.jsx
```diff
@@ -56,7 +75,8 @@ var Form = React.createClass({
     }
 
     return (
-     <label className={className} key={id} htmlFor={id}>
+     <label ref={id+'Label'} className={this.getInLabelClasses(id)} key={id} htmlFor={id}>
```
Pomax: I don't know what this name means. Can we make it `getLabelClasses`, since that's what it does?

templates/components/form/form.jsx
```diff
@@ -56,7 +75,8 @@ var Form = React.createClass({
...
+        {!this.isValid(id) ? errorTooltip : ''}
```
Pomax: we're calling `this.isValid(id)` a lot, even though we have a `dirty` state that tells us whether or not things are valid, right? There should really only ever be one call to `this.isValid()`, everything else we should be able to get from `this.state....`

templates/components/form/form.jsx
```diff
@@ -67,12 +87,21 @@ var Form = React.createClass({
      var fields = Object.keys(this.props.fields).map(this.buildFormElement);
      return <div role="form">{fields}</div>;
   },
-  getClasses: function(field) {
+  getInputClasses: function(field) {
     return React.addons.classSet({
       'form-control': true,
       'has-error': !this.isValid(field),
+      'is-valid': this.isValid(field)
```
Pomax: why would we call this twice? `var valid = this.isValid(field); ... return { 'has-error': !valid, 'is-valid': valid }` but if these states are only ever each other's opposite, why would we even have an `is-valid` class rather than just styling based on whether `has-error` is present or not?

templates/components/icontext.jsx
```diff
@@ -0,0 +1,21 @@
+var React = require('react');
```
Pomax: we should rename this to icontext.jsx, really

templates/less/components/form.less
Pomax: pretty sure you don't. This is a sign that the classes you're using are mutually exclusive and you're "pushing one of them to win the CSS war". is-valid and has-error should be mutually exclusive, so either the base grey gets overwritten with green, or it gets overwritten with red. We should not need `!important`

templates/less/components/form.less
```diff
@@ -76,3 +92,32 @@
     width: 400px;
   }
 }
+
+
+.warning:before {
```
Pomax: this is a LESS file, so just move this into warning:
```
.warning {
  ...
  &:before {
    ...
  }
}
```

templates/less/pages/signup.less
```diff
@@ -1,5 +1,11 @@
 .signup-page {
 
+  .btn.btn-awsm {
+    max-width: 165px;
+    float: right;
```
Pomax: why are we using floats?

templates/components/icontext.jsx
Pomax: looks like my editor was set to 4 spaces instead of 2

templates/pages/reset-password.jsx
```diff
@@ -1,10 +1,56 @@
...
+  },
   render: function() {
     return (
-      <div>Kill me please</div>
+      <div>
+        <Header redirectText="Need an account?" redirectLabel="Sign up" redirectPage="signup" />
+
+        <div>
+            {!this.state.submitForm && !this.state.email ?
+              <RequestView submitForm={this.handleResetPassword}/> : '' }
```
Pomax: `: false` means it won't even inject *anything*, rather than injecting an empty string which we're doing now

templates/pages/reset-password.jsx
```diff
@@ -1,10 +1,56 @@
...
+
+            {this.state.submitForm ?
+              <IconText
+              iconClass="emailSentIcon fa fa-envelope-o"
+              className="emailSent centerDiv">
+                <p className="emailSentHeader">Check your email</p>
+                <p>We&#39;ve emailed you instructions for creating a new password.</p>
+              </IconText> : ''}
+            {this.state.email ?
+              <ResetView username="" submitForm={this.handleRequestPassword}/> : ''
```
Pomax: `: false` to prevent any content injection

watch.js
```diff
@@ -1,18 +0,0 @@
-var spawn = require('child_process').spawn;
```
Pomax: hehehe

templates/components/form/form.jsx
```diff
@@ -25,41 +27,51 @@ var Form = React.createClass({
     return {
       username: null,
```
Pomax: these should still be "" instead of null.

templates/components/form/form.jsx
```diff
@@ -25,41 +27,51 @@ var Form = React.createClass({
...
     }
 
+    var errorTooltip = (
+      <span className="warning">{value.errorMessage}</span>
+    );
     return (
-     <label className={className} key={id} htmlFor={id}>
-        {label && labelPosition==='before' ? label : false}
+     <label ref={id+'Label'} className={this.getLabelClasses(id)} key={id} htmlFor={id}>
+        {!this.isValid(id) ? errorTooltip : ''}
```
Pomax: `: false`

templates/components/form/form.jsx
```diff
@@ -25,41 +27,51 @@ var Form = React.createClass({
...
       input = (<span>{input}<span/></span>);
     }
 
+    var errorTooltip = (
+      <span className="warning">{value.errorMessage}</span>
+    );
     return (
-     <label className={className} key={id} htmlFor={id}>
-        {label && labelPosition==='before' ? label : false}
+     <label ref={id+'Label'} className={this.getLabelClasses(id)} key={id} htmlFor={id}>
```
Pomax: do we need this ref? do we every use it?

templates/components/form/form.jsx
```diff
@@ -67,12 +79,23 @@ var Form = React.createClass({
...
+      'has-error': !isValid,
+      'is-valid': isValid
     });
   },
+  getLabelClasses: function(field) {
+    var classes = {};
+    var isValid = this.isValid(field);
+    var ref = this.refs[field + 'Input'];
+    classes[this.getIconClass(field)] = true;
+    classes[this.getIconClass('error')] = !isValid;
```
Pomax: `this.getIconClass('error')` and `this.getIconClass('valid')` are invariant. There's not much point in constantly looking them up, we should just compute these values during componentWillMount or something and then refer to those

templates/components/icontext.jsx
Pomax: still 4 spaces instead of 2?

templates/less/components/form.less
```diff
@@ -8,62 +8,100 @@
...
-    width: 100%;
-    font-size: 16px;
-    font-weight: 200;
-    height: 55px;
-    padding-left: 12.5%;
-    margin: 0 auto;
+
+  input[disabled] {
+    background-color: transparent;
+    cursor: default !important;
```
Pomax: still an `important` left here

Pomax: also, as an `input`, pretty sure this needs to go inside of `label` like the other inputs.

templates/less/components/form.less
```diff
@@ -8,62 +8,100 @@
...
+    right: 10px;
+    top: 17px;
+    color: #53C0A3;
   }
 }
-.has-error {
-  border: 1px solid #E95B95;
-  background-color: #FEEDED;
+
+input.has-error {
```
Pomax: move inside of label

templates/components/form/form.jsx
alicoding: for test I think

templates/less/components/form.less
```diff
@@ -76,3 +114,29 @@
     width: 400px;
   }
 }
+
+
+.warning {
```
Pomax: was warning independent, or does it need to be moved into label or input or something?

templates/less/pages/reset-password.less
```diff
@@ -0,0 +1,56 @@
+.requestPassword {
+  height: 180px;
+  padding-top: 26px;
+  .btn.btn-awsm {
+    max-width: 185px;
+  }
+}
+
+.resetPassword {
```
Pomax: the entire rest of this file should be *inside* the page's namespacing block. There should be nothing outside

templates/less/pages/reset-password.less
```diff
@@ -0,0 +1,56 @@
...
+  }
+  float: left;
+  display: inline-block;
+  vertical-align: bottom;
+  border: 2px solid #72AFDE;
+  margin-right: 1rem;
+  width: 7rem;
+  height: 7rem;
+  border-radius: 2.5em;
+}
```
Pomax: missing newline

templates/less/pages/signup.less
Pomax: still a float?

templates/pages/reset-password.jsx
```diff
@@ -1,10 +1,56 @@
...
+            {this.state.submitForm ?
+              <IconText
+              iconClass="emailSentIcon fa fa-envelope-o"
+              className="emailSent centerDiv">
+                <p className="emailSentHeader">Check your email</p>
+                <p>We&#39;ve emailed you instructions for creating a new password.</p>
+              </IconText> : false}
+            {this.state.email ?
+              <ResetView username="" submitForm={this.handleRequestPassword}/> : false
+            }
```
Pomax: in the previous instance `}` goes after `false`. Nit, but can we use the same pattern in L24 and here?

templates/pages/signup.jsx
```diff
@@ -47,8 +49,9 @@ var Signup = React.createClass({
         <h2>Free and open source – forever.</h2>
         <Form ref="userform" fields={fieldValues} validators={fieldValidators}/>
         <div className="commit">
-          <div className="agreement" />
-          <div className="eula">By signing up, I agree to Webmaker's <a href="" className="underline">Terms of Service</a> and <a href="" className="underline">Privacy Policy</a>.</div>
+          <IconText iconClass="agreement" textClass="eula">
+            By signing up, I agree to Webmaker's <a href="" className="underline">Terms of Service</a> and <a href="" className="underline">Privacy Policy</a>.
```
Pomax: just realized we're not actually pointing to anything here. Filed #82 on that

PR #81, opened 2015-03-27
change auth_code to code

cadecairos:


jbuck:
Since we need to start doing this, want to add Joi validation of the query parameter for the `code` parameter only?

web/server.js
```diff
@@ -116,6 +117,14 @@ module.exports = function(options) {
       method: 'GET',
       path: '/login/oauth/access_token',
       config: {
+        validate: {
+          query: {
+            code: Joi.string(),
+            client_id: Joi.string(),
+            client_secret: Joi.string(),
+            grant_type: Joi.string()
```
jbuck: I think this should be https://github.com/hapijs/joi#anyvalidvalue---aliases-only-equal right?

test/web.js
```diff
@@ -313,4 +313,70 @@ lab.experiment("OAuth", function() {
       });
     });
   });
+
+  lab.test("GET access_token - invalid grant_type", function(done) {
```
jbuck: I think these descriptions are wrong

test/web.js
```diff
@@ -313,4 +313,70 @@ lab.experiment("OAuth", function() {
...
+          "Content-Type": "application/x-www-form-urlencoded"
+        }
+      };
+
+      s.inject(authTokenRequest, function(authTokResponse) {
+        var redirectUri = url.parse(authTokResponse.headers.location, true);
+        accessTokenRequest.url += "&code=" + redirectUri.query.code;
+
+        s.inject(accessTokenRequest, function(response) {
+          Code.expect(response.statusCode).to.equal(400);
```
jbuck: I think it's also worth checking the error or message properties to make sure the correct error is being triggered

test/web.js
cadecairos: how?

jbuck: blargh, I put this on the wrong line

test/web.js
```diff
@@ -313,4 +313,70 @@ lab.experiment("OAuth", function() {
...
+          done();
+        });
+      });
+    });
+  });
+
+  lab.test("GET access_token - missing client_id", function(done) {
```
jbuck: I'm not sure of the value of testing two invalid error conditions here (missing client_id and wrong grant_type)

cadecairos: probably a copy paste error on my part

jbuck:
This PR needs to be re-done against develop

PR #87, opened 2015-03-27
Stop testing Node v0.10

jbuck:


jbuck:
@alicoding @cadecairos @gvn @k88hudson @Pomax @simonwex any objections to not caring about node v0.10?

Pomax:
:boom:

k88hudson:
Hm, actually isn't it recommended that we stick to node v0.10? I have found 0.12 to be somewhat buggy, especially with older code...

simonwex:
Buggy how?

jbuck:
I already merged it, but I'm happy to back it out too. I guess what this patch does here is just remove CI for node v0.10. I hadn't even considered using node v0.10 for staging, I was planning on using node 0.12 for staging.

I looked for some recommendations, and Eran Hammer, former head of Walmart Mobile says [0.10 or iojs](http://hueniverse.com/2015/03/02/the-node-version-dilemma/).

Pomax:
one of the major wins here is npm 2.5 -- but that said, I'd also like to know where 0.12 is still considered buggy

toolness:
I'm not sure what k88 means by "buggy" but I have run into problems w/ running slightly older code on 0.12 due to API breakages in some of the experimental core node APIs and such, e.g. https://github.com/plus3network/gulp-less/issues/140. In the case of the Teach site it's actually preferable for us to use a slightly older version of LESS but we can't because it doesn't support 0.12 and, um, my computer has 0.12 on it ... and I'm lazy I guess?

Pomax:
the older version of less is pretty behind at this point though, 3.x should be 0.12 compatible and have a number of fixes compared to the 2.x version

Issue #53, opened 2015-03-24
Implement user migration page


alicoding:
See #28 for designs.

cc @secretrobotron @cassiemc

Pomax:
Final screens from the above ticket:

![01-enter-username](https://cloud.githubusercontent.com/assets/1186326/6829903/d73d2c78-d2ec-11e4-8e7e-01f6c60d77d5.jpg)
![02-enter-token](https://cloud.githubusercontent.com/assets/1186326/6829904/d73f3e96-d2ec-11e4-9521-2cf3f6815e72.jpg)
![03-set-password](https://cloud.githubusercontent.com/assets/1186326/6829905/d7415a78-d2ec-11e4-85b3-770aed1cd902.jpg)
![04-success-v2](https://cloud.githubusercontent.com/assets/1186326/6829906/d74173e6-d2ec-11e4-9e02-a6ad462cdcf0.jpg)

secretrobotron:
@cadecairos is working on a good flow for this at the moment, which will let users click on a link in an email to hop right to the password reset screen.

secretrobotron:
@alicoding @cadecairos can this close now?

alicoding:
I think so. Right now I'm only working on the UX for many different cases but that we already filed them in different tickets.

PR #88, opened 2015-03-27
User migration form

alicoding:
Ref to https://github.com/mozilla/id.webmaker.org/issues/53

### DIFF
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -12,7 +12,8 @@ var Form = React.createClass({
         'password': 'icon-label-password',
         'email':    'icon-label-email',
         'error':    'icon-label-error',
-        'valid':    'icon-label-valid'
+        'valid':    'icon-label-valid',
+        'key':      'icon-label-password'
       }
   },
   mixins: [
@@ -31,7 +32,8 @@ var Form = React.createClass({
       password: '',
       email: '',
       checked: false,
-      dirty: {}
+      dirty: {},
+      key: ''
     };
   },
   dirty: function(id) {
@@ -83,7 +85,6 @@ var Form = React.createClass({
   getInputClasses: function(field) {
     var isValid = this.isValid(field);
     return React.addons.classSet({
-      'form-control': true,
       'has-error': !isValid,
       'is-valid': isValid
     });
@@ -92,6 +93,7 @@ var Form = React.createClass({
     var classes = {};
     var isValid = this.isValid(field);
     var ref = this.refs[field + 'Input'];
+    classes['inputBox'] = field === 'feedback';
     classes[this.getIconClass(field)] = true;
     classes[this.errorClass] = !isValid;
     classes[this.validClass] = (this.state.dirty[field] && isValid) || this.passChecked
--- a/templates/components/login-no-pass-form.jsx
+++ b/templates/components/login-no-pass-form.jsx
@@ -15,12 +15,12 @@ var fields = [
 
 var fieldsValidators = validators.getValidatorSet(fields);
 
-var RequestResetPassword = React.createClass({
+var LoginNoPassword = React.createClass({
   render: function() {
     return (
-      <div className="requestPassword innerForm centerDiv">
+      <div className="loginNoPass innerForm centerDiv">
         <Form ref="userform" fields={fields} validators={fieldsValidators} />
-        <button onClick={this.processFormData} className="btn btn-awsm">Set a new password</button>
+        <button onClick={this.processFormData} className="btn btn-awsm">Log in</button>
       </div>
     );
   },
@@ -30,4 +30,4 @@ var RequestResetPassword = React.createClass({
   }
 });
 
-module.exports = RequestResetPassword;
+module.exports = LoginNoPassword;
--- a/templates/components/migrate-key-form.jsx
+++ b/templates/components/migrate-key-form.jsx
@@ -1,0 +1,43 @@
+var React = require('react');
+var validators = require('../lib/validatorset');
+var Form = require('./form/form.jsx');
+var IconText = require('./icontext.jsx');
+
+var fields = [
+  {
+    'key': {
+      'placeholder': 'Paste your key here',
+      'type': 'password',
+      'validator': 'key',
+      'errorMessage': 'Invalid key'
+    }
+  }
+];
+
+var fieldsValidators = validators.getValidatorSet(fields);
+
+var MigrateKey = React.createClass({
+  render: function() {
+    return (
+      <div className="migrateKeyContainer">
+        <IconText
+          iconClass="emailSentIcon fa fa-envelope-o"
+          className="emailSent arrow_box"
+          header="Check your email"
+          headerClass="emailSentHeader">
+            <p>We've emailed your login key to the address you provided.</p>
+        </IconText>
+        <div className="migrateKey innerForm">
+          <Form ref="userform" fields={fields} validators={fieldsValidators} />
+          <button onClick={this.processFormData} className="btn btn-awsm">Login</button>
+        </div>
+      </div>
+    );
+  },
+  processFormData: function() {
+    var form = this.refs.userform;
+    form.processFormData(this.props.submitForm);
+  }
+});
+
+module.exports = MigrateKey;
--- a/templates/components/set-password-migration-form.jsx
+++ b/templates/components/set-password-migration-form.jsx
@@ -1,0 +1,43 @@
+var React = require('react');
+var validators = require('../lib/validatorset');
+var Form = require('./form/form.jsx');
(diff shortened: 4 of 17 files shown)


templates/components/form/form.jsx
```diff
@@ -12,7 +12,8 @@ var Form = React.createClass({
         'password': 'icon-label-password',
         'email':    'icon-label-email',
         'error':    'icon-label-error',
-        'valid':    'icon-label-valid'
+        'valid':    'icon-label-valid',
+        'key':    'icon-label-password'
```
Pomax: spacing~

templates/components/login-no-pass-form.jsx
```diff
@@ -0,0 +1,33 @@
...
+      </div>
+    );
+  },
+  processFormData: function() {
+    var form = this.refs.userform;
+    form.processFormData(this.props.submitForm);
+  }
+});
+
+module.exports = LoginNoPassword;
```
Pomax: newline

templates/components/set-password-migration-form.jsx
```diff
@@ -0,0 +1,42 @@
...
+var fieldsValidators = validators.getValidatorSet(fields);
+
+var SetPasswordMigration = React.createClass({
+  render: function() {
+    return (
+      <div className="migrateKeyContainer">
+        <IconText
+          iconClass="emailSentIcon fa fa-lock"
+          className="emailSent arrow_box">
+            <p className="emailSentHeader">Set your password</p>
```
Pomax: the IconText element has a special attribute for headers, so do we want a `<p>` here, or should this be the proper header?

alicoding: It needs to be a proper header where I have control over the line height and spacing. Not sure if the IconText is uses the same... I have tried the other day and thought each page doesn't seem to have the same layout and I ended up using this.

templates/less/components/form.less
```diff
@@ -1,20 +1,4 @@
...
-  @media (min-width : 451px) {
-    width: 400px;
-    height: 250px;
-  }
-  border-radius: 7px;
-  padding-top: 25px;
-  margin-bottom: 30px;
-  background-color: white;
-
+form {
```
Pomax: this feelds wrong. Why `form` instead of `.form`? Are we using forms again instead of divs?

templates/less/pages/signup.less
```diff
@@ -32,7 +34,7 @@
 
   div[role=form] {
     background: @white;
-    padding: 4rem;
+    padding: 25px;
```
Pomax: this is a thing that'll need to be reviewed soon, but rem for things that are dependent on text is generally better than pixels

templates/less/styles.less
```diff
@@ -41,3 +42,12 @@ h3 {
...
 }
+
+
+.innerForm {
+  padding: 25px;
+  text-align: center;
+  @media (min-width : 451px) {
+    text-align: right;
+  }
+}
```
Pomax: newline

templates/lib/routes.jsx
```diff
@@ -9,6 +9,7 @@ var routes = (
     <Route name="reset-password" path="/reset-password/?" handler={require('../pages/reset-password.jsx')}/>
     <Route name="login"          path="/login/?"          handler={require('../pages/login.jsx')}/>
     <Route name="signup"          path="/signup/?"          handler={require('../pages/signup.jsx')}/>
```
Pomax: look like there's some spacing-creep

templates/lib/validatorset.js
```diff
@@ -4,7 +4,8 @@ var fieldValidators = {
   username:       Joi.string().alphanum().min(1).required().label('Username'),
   password:       Joi.string().min(8).required().label('Password'),
   verifyPassword: Joi.any().valid(Joi.ref('password')).required().label('Password Confirmation'),
-  email:          Joi.string().email().required().label('Email')
+  email:          Joi.string().email().required().label('Email'),
+  key:      Joi.string().alphanum().min(1).required().label('Key')
```
Pomax: spacing

templates/pages/migrate.jsx
```diff
@@ -0,0 +1,71 @@
...
+  getInitialState: function() {
+    return {
+      login: false,
+      setKey: false,
+      setPass: false,
+      success: false
+    };
+  },
+  render: function() {
+
```
Pomax: rogue empty lines

templates/pages/migrate.jsx
```diff
@@ -0,0 +1,71 @@
...
+    var content = (<LoginNoPasswordForm submitForm={this.handleLogin}/>);
+    if(this.state.setKey) {
+      content = (<MigrateKeyForm submitForm={this.handleSetKey} />);
+    } else if(this.state.setPass) {
+      content = (<SetPasswordMigrationForm submitForm={this.handleResetPassword} />);
+    } else if(this.state.success) {
+      content = (<IconText
+          iconClass="successBannerIcon icon fa fa-check"
+          className="successBanner centerDiv">
+            <p className="successBannerHeader">Success!</p>
```
Pomax: should this be a real header in the IconText element or a paragraph?

templates/pages/reset-password.jsx
```diff
@@ -19,17 +19,17 @@ var ResetPassword = React.createClass({
...
 
             {this.state.submitForm ?
               <IconText
-              iconClass="emailSentIcon fa fa-envelope-o"
-              className="emailSent centerDiv">
-                <p className="emailSentHeader">Check your email</p>
-                <p>We&#39;ve emailed you instructions for creating a new password.</p>
+                iconClass="emailSentIcon fa fa-envelope-o"
+                className="emailSent centerDiv">
+                  <p className="emailSentHeader">Check your email</p>
```
Pomax: should this be a header, or a paragraph

templates/pages/reset-password.jsx
```diff
@@ -19,17 +19,17 @@ var ResetPassword = React.createClass({
...
+                  <p>We&#39;ve emailed you instructions for creating a new password.</p>
```
Pomax: why would you ever escape a quote? This never needed htmlentities, not even in 1995...

templates/components/form/form.jsx
```diff
@@ -78,12 +80,11 @@ var Form = React.createClass({
   },
   render: function() {
      var fields = Object.keys(this.props.fields).map(this.buildFormElement);
-     return <div role="form">{fields}</div>;
+     return <div role="form"><form>{fields}</form></div>;
```
Pomax: why is this a real HTML form all of a sudden?

alicoding: I just though doing this might be useful in term of applying style to every place as a standard.

templates/less/components/form.less
alicoding: Its the same for me I can change it if there is a good reason why we have to change

templates/components/migrate-key-form.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+var MigrateKey = React.createClass({
+  render: function() {
+    return (
+      <div className="migrateKeyContainer">
+        <IconText
+          iconClass="emailSentIcon fa fa-envelope-o"
+          className="emailSent arrow_box">
+            <p className="emailSentHeader">Check your email</p>
+            <p>We&#39;ve emailed your login key to the address you provided.</p>
```
Pomax: no need to use an htmlentity here

templates/components/set-password-migration-form.jsx
Pomax: it's a header, you style it as needed based on 
```
.pagename {
  .icontext {
    h1 {
      /* stuff here */
    }
  }
}
```

templates/components/form/form.jsx
Pomax: we already have div[role=form]. Turning it into a magically active HTML element is probably a really bad plan

templates/less/components/form.less
Pomax: yeah: don't use a form unless it's *actually* a form. We're not using real forms, so let's not use a tag that adds lots of semantic interpretation. We already have `div[role=form]` to select with

templates/components/migrate-key-form.jsx
```diff
@@ -0,0 +1,42 @@
...
+var fieldsValidators = validators.getValidatorSet(fields);
+
+var MigrateKey = React.createClass({
+  render: function() {
+    return (
+      <div className="migrateKeyContainer">
+        <IconText
+          iconClass="emailSentIcon fa fa-envelope-o"
+          className="emailSent arrow_box">
+            <p className="emailSentHeader">Check your email</p>
```
Pomax: is this not a header, too?

PR #89, opened 2015-03-27
Reorganize package.json scripts

jbuck:


jbuck:
@gvn r?

The general idea here was:

* put most-used scripts at the top (start and test)
* group and sort everything else somewhat alphabetically

package.json
```diff
@@ -44,21 +51,24 @@
...
     "hoek": "^2.11.1",
     "hyperquest": "^1.0.1",
+    "imports-loader": "^0.6.3",
     "joi": "^6.0.8",
+    "jsx-loader": "^0.12.2",
     "react": "^0.12.2",
     "react-router": "^0.12.4",
+    "react-router-stub": "0.0.5",
     "react-validation-mixin": "^4.0.3",
+    "should": "^5.2.0",
```
gvn: Dev dep?

package.json
```diff
@@ -4,19 +4,26 @@
...
-    "build": "npm run build:testassets && npm run build:js",
-    "watch-css": "autoless --source-map templates/less public",
-    "watch-js": "npm run build:js -- --watch",
-    "jshint": "jshint  lib test web *.js",
+
     "jscs": "jscs lib test web *.js",
-    "js-beautify": "js-beautify  lib test web *.js -r ",
-    "validate": "npm run jshint && npm run jscs"
+    "jshint": "jshint lib test web *.js",
+    "js-beautify": "js-beautify lib test web *.js -r",
```
gvn: Maybe `jsbeautify` instead? I dunno...

gvn:
LGTM...just a few comments that I'll leave up to you, but R+!

package.json
jbuck: will fix

package.json
jbuck: sadly not, since the build step *also* builds our JS files it's actually a `build:js` dependency

PR #91, opened 2015-03-30
Loginapi

cadecairos:


cadecairos:
This PR needs review still, but if @alicoding wants to get a head start on hooking up the front end, go for it! otherwise I'll do that tomorrow. the loginAPI server will have to be running this commit or later: https://github.com/mozilla/login.webmaker.org/commit/117c841089951c5b9ef6a47401a3eed2739f4a51

lib/account.js
```diff
@@ -27,32 +56,94 @@ module.exports = function(config) {
...
+        });
+      });
+
+      resetRequest.end(JSON.stringify({
+        user: {
+          email: request.payload.email,
+          username: request.payload.username,
+          password: request.payload.password,
+          mailingList: request.payload.feedback,
+          prefLocale: request.payload.lang
```
alicoding: are we expecting this to be pass in somewhere from the client? or we are just hardcoding the value for now? Since the UI doesn't support that currently.

templates/pages/reset-password.jsx
```diff
@@ -30,22 +41,79 @@ var ResetPassword = React.createClass({
                 className="emailSent centerDiv"
                 headerClass="emailSentHeader"
                 header="Check your email">
-                  <p>We've emailed you instructions for creating a new password.</p>
+                  <p>We&rsquo;ve emailed you instructions for creating a new password.</p>
```
alicoding: me likey!

templates/pages/reset-password.jsx
```diff
@@ -6,18 +6,29 @@ var ResetView = require('../components/reset-password-view.jsx');
...
 var Router = require('react-router');
 
+
+var Url = require('url');
+require('whatwg-fetch');
+
 // This wraps every view
 var ResetPassword = React.createClass({
   getInitialState: function() {
+    var queryObj = Url.parse(window.location.href, true).query;
```
alicoding: This will be problematic when we do test with no `window` context, but we can deal with this later when we add the test.

templates/pages/reset-password.jsx
```diff
@@ -30,22 +41,79 @@ var ResetPassword = React.createClass({
...
+      },
+      body: JSON.stringify({
+        uid: data.username,
+        password: data.password,
+        resetCode: this.state.queryObj.resetCode
+      })
+    }).then(function(response) {
+      var queryObj,
+        redirectObj;
+      if ( response.status === 200 ) {
```
alicoding: is there any other possible response.status code that we want to handle here? or that's for later?

web/server.js
```diff
@@ -2,8 +2,8 @@ var Boom = require('boom');
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Joi = require('joi');
-var OAuthDB = require('../lib/oauth-db');
 var Path = require('path');
+var OAuthDB = require('../lib/oauth-db');
```
jbuck: This shouldn't be moved

web/index.js
```diff
@@ -6,7 +6,8 @@ var options = {
   loginAPI: process.env.LOGINAPI,
   oauth_clients: process.env.OAUTH_DB ? JSON.parse(process.env.OAUTH_DB) : [],
   cookieSecret: process.env.COOKIE_SECRET,
-  secureCookies: process.env.SECURE_COOKIES === 'true'
+  secureCookies: process.env.SECURE_COOKIES === 'true',
+  uri: process.env.URI
```
alicoding: can you document this as well?

lib/account.js
cadecairos: It can come from anywhere, really.

templates/pages/reset-password.jsx
cadecairos: error states can be handled later, since there currently aren't any

cadecairos:
@jbuck fixed up, r?

jbuck:
@cadecairos r+, just be careful of the merge conflict

Issue #77, opened 2015-03-26
Metrics: Requirements for Analytics and Database


adamlofting:
I'm closing the requirements gathering checklist https://github.com/mozilla/id.webmaker.org/issues/56 as we need to move on implementation. 

## Known tasks:

For @adamlofting and or dev support (TBC @simonwex)

- [x] Setup a new GA property via MoCo for id.webmaker.org ([Bug is here](https://bugzilla.mozilla.org/show_bug.cgi?id=1149508))
- [x] Setup basic Google Analytics in the id.webmaker.org codebase ([PR is open here #93](https://github.com/mozilla/id.webmaker.org/pull/93))
  * Use [react-ga](https://github.com/adamlofting/react-ga) module to do this
- [x] Review the front-end experience and tech stack to see what tracking beyond basic pageviews is required
- [x] Setup GA tracking of the in-page event tracking we care about (validation errors, field focus, etc) #137 
- Setup conversions Goals and funnels in the GA account
  - [x] Account creation and login
  - [ ] Field by field funnel
- [x] Ensure we're tracking pageload times as part of our GA configuration
- [x] Solve / Spec how we identify traffic sources (teach / webmaker / other) within the GA reporting (this is covered by standard GA referral data) 
- [ ] Solve / Spec how we update GA code, and reporting configuration across all other properties currently using login
- [x] Not a blocker: Proof of concept implementation for Optimizely

## Open Questions:

For @jbuck, @secretrobotron, or others in the know.

- [x] Are we replacing the existing `Users` table that sits behind the current login platform?
  * This shouldn't block or slow down shipping, but will cause a bunch of knock-on work for our integration db and KPI dashboards
- [x] Are we launching just with teach, or swapping out all instances of login at the same time?
- [x] Did we preserve the existing work on the [Referrer API](https://wiki.mozilla.org/Webmaker/Maker_Party/referrer_api) and [this](https://github.com/mozilla/login.webmaker.org/commit/b87b38ce308d84e347d1742584a0c7bd57413953), or has this been killed? Solved in #97
- [x] How will we identify in the database, the difference between Webmaker Account Holders, and Webmaker Users? We need some kind of table or flag to say which sites the user has authenticated against. Example query I'd want to be able to run = `show me all account holders who use teach.m.o but never webmaker.org`. Solved in #97

## Reporting Requirements: 

For @adamlofting

### For ID as a standalone Service

Conversion rates
- [x] UV to Sign Up 
[...]

PR #93, opened 2015-03-31
Add Basic Google Analytics

adamlofting:
This is the first of many metrics requirements, as recorded in https://github.com/mozilla/id.webmaker.org/issues/77

Adds the `react-ga` module, and implements basic pageview tracking using `react-router`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,22 +6,17 @@
   "scripts": {
     "start": "parallelshell \"node web/index.js\" \"npm run watch:css\" \"npm run watch:js\"",
     "test": "npm run test:server && npm run test:browser && npm run lint",
-
     "build": "npm run build:css && npm run build:js",
     "build:css": "npm run watch:css -- --no-watch",
     "build:js": "webpack --config webpack.config.js --progress --profile --colors",
-
     "jscs": "jscs lib test web *.js",
     "jshint": "jshint lib test web *.js",
     "jsbeautify": "js-beautify lib test web *.js -r",
     "lint": "npm run jshint && npm run jscs",
-
     "postinstall": "npm run build",
-
     "pretest:browser": "ncp node_modules/mocha/mocha.js public/tests/mocha.js && ncp node_modules/mocha/mocha.css public/tests/mocha.css",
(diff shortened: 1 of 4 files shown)


adamlofting:
r?

@jbuck let me know if I should assign these tickets to anyone else to review?

templates/lib/routes.jsx
```diff
@@ -3,6 +3,11 @@ var Router = require('react-router');
 var Route = Router.Route;
 var Link = Router.Link;
 var NotFoundRoute = Router.NotFoundRoute;
+var ga = require('react-ga');
+
+// TODO: this is a dummy GA tracking ID until the real one is ready
+var gaTrackingID = process.env.GA_TRACKING_ID || 'UA-59356678-5';
```
alicoding: You need to add `GA_TRACKING_ID` and `GA_DEBUG` in the webpack.config.js file in the plugin to expose them client side as well.
``` js
  plugins: [
    new webpack.DefinePlugin(importEnvVars([
      'GA_TRACKING_ID',
      'GA_DEBUG'
    ]))
  ]
```

package.json
```diff
@@ -6,22 +6,17 @@
   "scripts": {
     "start": "parallelshell \"node web/index.js\" \"npm run watch:css\" \"npm run watch:js\"",
     "test": "npm run test:server && npm run test:browser && npm run lint",
-
```
alicoding: Not sure if these lines removal are intentional?

alicoding:
Just two things to fix otherwise R+ from me.

package.json
adamlofting: I think this happened when I ran `npm install react-ga --save`

I'm guessing this is going to be a recurring issue if we want to keep them formatted like this?

templates/lib/routes.jsx
adamlofting: Will do, and good to know. Also, I should add them to the docs. :+1:

package.json
alicoding: I don't mind tbh, but not sure what was the purpose really... @jbuck ?

templates/lib/routes.jsx
adamlofting: These have been added now :smile:

PR #94, opened 2015-03-31
Manual tests

k88hudson:


alicoding:
hmm do we want to commit these two files?

` public/manual.bundle.js` and ` public/manual.bundle.js.map`?

alicoding:
r+ with removing that two files and add them as part of `.gitignore`?

k88hudson:
oh right! good catch

k88hudson:
@alicoding what do you think about ignoring `*.bundle.*`

alicoding:
I like that approach good thinking! r+!

PR #99, opened 2015-04-01
Make 'npm test' work on Windows by gulp-ifying jshint and jscs tasks.

### LINKED ISSUE #98: npm test fails on windows during jshint stage
Here is where it explodes:

```
> id.webmaker.org@1.0.0 lint c:\Users\Atul\Documents\mozilla\id.webmaker.org
> npm run jshint && npm run jscs


> id.webmaker.org@1.0.0 jshint c:\Users\Atul\Documents\mozilla\id.webmaker.org
> jshint lib test web *.js

ERROR: Can't open *.js
ERROR: Can't parse config file: c:\Users\Atul\Documents\mozilla\id.webmaker.org\.jshintrc
Error:SyntaxError: Unexpected token o
```

I'm using Windows 7 and this seems to happen with both `cmd.exe` and the `bash` that comes with Git for Windows.

toolness:
This fixes #98 by making `jshint` and `jscs` into gulp tasks, just like they are on the Teach site.

(Sorry, I didn't realize that this project didn't already use gulp until I was half done... I only realized way too late that the gulpfile I was editing was actually mofo-style's... Guess I wasn't paying enough attention when I was using Sublime Text's "Goto Anything" feature!)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.js text eol=lf
```
toolness: This is a workaround for https://github.com/MozillaFoundation/javascript-style-guide/issues/13.

toolness:
This PR makes `npm test` work on my Windows machine, but I haven't tried it out yet on OSX/Linux, so we should definitely pay attention to what Travis thinks here. :)

jbuck:
Is it really not possible to specify a config file without using Gulp?

toolness:
It probably is, but I suspect we'd still run into the `ERROR: Can't open *.js` problem mentioned in #98? Gulp seems to deal with globs in a more platform-independent way than `npm` at least...

toolness:
Yeah, I just tried redefining `jshint` to be gulp-less in my `package.json`, and though it finds the config file, it still fails on Windows (`bash` *and* `cmd.exe`, doesn't matter which):

```
> jshint -c node_modules/mofo-style/linters/.jshintrc lib test web *.js

ERROR: Can't open *.js
```

jbuck:
The only `.js` file we have in the tree at the moment is `webpack.config.js`. We would just modify the script to refer to that file directly.

toolness:
Er wait would just passing `.` instead of `*.js` or a literal list of files also work? My only concern w/ passing specific filenames is that the list might not get updated as new files are added to the root dir (though obviously that won't happen often)...

toolness:
Hmm passing `.` didn't work so I will just name the files explicitly.

Closing this PR, we'll use the de-gulped #100 instead.

PR #100, opened 2015-04-01
Make jscs and jshint work on windows w/o gulp.

### LINKED ISSUE #98: npm test fails on windows during jshint stage

toolness:
fixes #98.


.gitattributes
```diff
@@ -0,0 +1 @@
+*.js text eol=lf
```
toolness: This is a workaround for https://github.com/MozillaFoundation/javascript-style-guide/issues/13.

jbuck:
@toolness This looks good! When travis goes green... r+

package.json
```diff
@@ -9,8 +9,8 @@
     "build": "npm run build:css && npm run build:js",
     "build:css": "npm run watch:css -- --no-watch",
     "build:js": "webpack --config webpack.config.js --progress --profile --colors",
-    "jscs": "jscs lib test web *.js",
-    "jshint": "jshint lib test web *.js",
+    "jscs": "jscs -c node_modules/mofo-style/linters/.jscsrc lib test web webpack.config.js",
+    "jshint": "jshint -c node_modules/mofo-style/linters/.jshintrc lib test web webpack.config.js",
```
toolness: I am not a big fan of the ol' [DRY violation](http://en.wikipedia.org/wiki/Don%27t_repeat_yourself) here but I guess it's all we can do if we don't want to use gulp or a similar JS thingey.

PR #117, opened 2015-04-01
Fix OAUTH_DB example value in README

toolness:
It seems the current implementation is actually looking at the `client_secret` key, not the `secret` key.

I also recommend doing some basic validation in the `oauth-db.js` file to ensure that the JSON blob is well-formed, to avoid having to debug things like this further down the pipeline.


secretrobotron:
@jbuck R?

jbuck:
@toolness r+

PR #121, opened 2015-04-02
Change username field to username or email

### LINKED ISSUE #120: Change "username" field to "username or email"
I noticed that the login page has a "username" field:

Based on user testing with the Hive Directory--as well as personal experience--I've noticed that users easily forget their username if they don't frequently use the site (or if the site doesn't constantly use the username in a memorable way, e.g. through social functionality). As such, I recommend doing what a lot of other sites (including GitHub and the [Hive Directory](https://directory.hivenyc.org/accounts/login/)) do, and changing the "username" field to "username or email".

alicoding:
Closes #120

### DIFF
--- a/templates/lib/validatorset.js
+++ b/templates/lib/validatorset.js
@@ -1,8 +1,11 @@
 var Joi = require('joi');
 
 var fieldValidators = {
-  username:       Joi.string().alphanum().min(1).required().label('Username'),
-  password:       Joi.string().min(8).required().label('Password'),
+  username:       Joi.alternatives().try(
+                    Joi.string().alphanum().min(1).required(),
+                    Joi.string().email().required())
+                  .label('Username or email'),
+  password:       Joi.string().regex(/^\S{8,16}$/).label('Password'),
   verifyPassword: Joi.any().valid(Joi.ref('password')).required().label('Password Confirmation'),
   email:          Joi.string().email().required().label('Email'),
   key:            Joi.string().alphanum().min(1).required().label('Key')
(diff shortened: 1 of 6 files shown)


alicoding:
@toolness R?

alicoding:
Since we are accepting username or email in that field anyway so might as well.

templates/lib/validatorset.js
```diff
@@ -1,8 +1,11 @@
 var Joi = require('joi');
 
 var fieldValidators = {
-  username:       Joi.string().alphanum().min(1).required().label('Username'),
-  password:       Joi.string().min(8).required().label('Password'),
+  username:       Joi.alternatives().try(
+                    Joi.string().alphanum().min(1).required(),
+                    Joi.string().email().required())
+                  .label('Username or email'),
+  password:       Joi.string().regex(/^\S{8,16}$/).label('Password'),
```
toolness: Because I don't really know anything about this `joi` library, I don't feel particularly qualified to R+ this part... And since I don't actually know the context of the rest of the codebase I'm not sure if I should R+ either? Maybe @jbuck or @cadecairos or someone else w/ a better understanding of it can review?

toolness: Thanks a lot for doing this though, I really appreciate it :)

alicoding: No problem @toolness ! I will flag @cadecairos for R? then :+1:

jbuck:
@thisandagain I think we wanted to use username only, right?

toolness:
If we actually *say* username but *accept* username or email, that might actually be fine. I know with the Hive Directory folks were actually entering their email in the username field, but because I hadn't specifically coded it to accept emails, the login would always fail for them when they tried it.

thisandagain:
We might want to detect if an email address is provided and have a friendly prompt along the lines of, "Oops! That looks like an email address. Please enter your username."

I put the reasoning for why we don't want to do this in #120

PR #125, opened 2015-04-02
logout route and test

### LINKED ISSUE #119: Document and/or implement logout API
I'm noticing as I use the server that it automatically calls my OAuth2 callback (without prompting the user) if the user is already logged-in. This seems to imply that there should be some kind of logout endpoint that I can ping in order to explicitly log *out* of the id server--without such a feature, for instance, there seems to be no way for one user to log out of the teach website and then hand their device to another human, who can then log in as a different user. (This isn't just an edge case, it's also a security hazard, as there's effectively no way to log out of the system--upon clicking 'login', the second user would immediately be logged in as the first user without any prompting.)

In fact, it seems the only way I can forcibly "log out" of the system right now is to change the server's cookie secret, which invalidates my sesssion cookie! lol. So implementing and/or documenting a logout endpoint would be very helpful here.

cadecairos:
Fixes #119

### DIFF
--- a/test/web.js
+++ b/test/web.js
@@ -331,13 +331,40 @@ lab.experiment("OAuth", function() {
     });
   });
 
-  lab.test("POST logout - no client_id", function(done) {
+  lab.test("GET logout", function(done) {
     var request = {
-      method: "POST",
+      method: "GET",
+      url: "/logout?client_id=test",
+      credentials: {
+        uid: "webmaker"
+      }
+    };
+
+    s.inject(request, function(response) {
+      Code.expect(response.statusCode).to.equal(302);
+      Code.expect(response.headers["set-cookie"][0]).to.equal(
+        "webmaker=; Max-Age=0; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Secure; HttpOnly; Path=/"
+      );
+
+      Code.expect(response.headers.location).to.exist();
+
+      var redirectUri = url.parse(response.headers.location, true);
+
+      Code.expect(redirectUri.protocol).to.equal("http:");
+      Code.expect(redirectUri.host).to.equal("example.org");
+      Code.expect(redirectUri.pathname).to.equal("/oauth_redirect");
+      Code.expect(redirectUri.query.logout).to.equal("true");
+
+      ls.stop(done);
+    });
+  });
+
+  lab.test("GET logout - no client_id", function(done) {
+    var request = {
+      method: "GET",
       url: "/logout",
-      payload: {
-        uid: "webmaker",
-        password: "notThePassword"
+      credentials: {
+        uid: "webmaker"
       }
     };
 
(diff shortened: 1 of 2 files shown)


web/server.js
```diff
@@ -261,6 +261,17 @@ module.exports = function(options) {
           reply(json.user);
         });
       }
+    },
+    {
+      method: 'POST',
+      path: '/logout',
+      config: {
+        auth: false
```
jbuck: Don't we want this to require authentication?

toolness: Authentication schmauthentication!!

Actually I have noticed that a lot of web frameworks seem to have an auth-less `GET` endpoint for logout, which is quite counterintuitive. But if you did this, or at least kept the POST endpoint auth-less and csrf-exempt, it would make it possible for us to play around with having logging out of the teach site *also* log the user out of id.wm.o, which could improve UX and security (see #119 for more discussion on that).

toolness: I suspect part of the reason that some web frameworks put so little security around logout endpoints is that logging out generally always results in the same or less privileges than the user had before, so it's never a threat, and should always be easy to do? I mean, at most it seems like you could annoy someone by constantly logging them out of a third-party site without their permission, but that seems like the extent of the damage?

I have no idea.

jbuck: Idea: what if we had this `/logout` endpoint become a GET *but* also require a **client_id** parameter. When you logout, it'll redirect you to the logout URL?

I'm not certain that even having a separate logout URL is a good idea... I guess it does depend on whether login state should be considered global or not.

web/server.js
```diff
@@ -261,6 +261,17 @@ module.exports = function(options) {
           reply(json.user);
         });
       }
+    },
+    {
+      method: 'POST',
```
jbuck: change this to a GET

web/server.js
toolness: If we do that can we still have it communicate the `state` back to the URL? In fact, if it communicates `state`, we don't even have to have a separate logout URL... we can just call the callback URL, and the client (i.e. teach site) can be responsible for setting state in such a way that it knows the callback was called in response to a logout, and any other necessary stuff it may need.

cadecairos:
@jbuck r?

test/web.js
```diff
@@ -291,9 +291,9 @@ lab.experiment("OAuth", function() {
       var request = {
         method: "POST",
         url: "/login",
-        payload: {
+        credentials: {
```
jbuck: I don't think this change is correct here?

test/web.js
```diff
@@ -331,6 +331,34 @@ lab.experiment("OAuth", function() {
     });
   });
 
+  lab.test("POST logout - no client_id", function(done) {
+    var request = {
+      method: "POST",
```
jbuck: This should be a GET request

test/web.js
```diff
@@ -331,6 +331,34 @@ lab.experiment("OAuth", function() {
...
+      url: "/logout",
+      payload: {
```
jbuck: There should be no payload here

web/server.js
```diff
@@ -261,6 +261,36 @@ module.exports = function(options) {
...
+      path: '/logout',
+      config: {
+        auth: false,
+        pre: [
+          {
+            assign: 'redirectUri',
+            method: function(request, reply) {
+              if ( !request.query.client_id ) {
+                return reply('https://webmaker.org/?logout=true');
```
jbuck: This just needs to be `https://webmaker.org` right?

jbuck:
@cadecairos r+ on a green travis build

PR #128, opened 2015-04-02
get user data

cadecairos:
This isn't done yet, still needs testing.

### DIFF
--- a/lib/scopes.js
+++ b/lib/scopes.js
@@ -1,0 +1,29 @@
+var SCOPES = {
+  user: [
+    'username'
+  ],
+  email: [
+    'email'
+  ]
+};
+
+module.exports = {
+  filterUserForScopes: function(user, scopes) {
+    var filtered = {};
+
+    if ( !Array.isArray(scopes) ) {
+      scopes = [scopes];
+    }
+
+    scopes.forEach(function(scope) {
+      var scopeAttrs = SCOPES[scope];
+      if ( scopeAttrs ) {
+        scopeAttrs.forEach(function(attr) {
+          filtered[attr] = user[attr];
+        });
+      }
+    });
+
+    return filtered;
+  }
+};
--- a/lib/scopes.js
+++ b/lib/scopes.js
@@ -1,6 +1,8 @@
 var SCOPES = {
   user: [
-    'username'
+    'username',
+    'avatar',
+    'id'
   ],
   email: [
     'email'
--- a/test/web.js
+++ b/test/web.js
@@ -636,7 +636,7 @@ lab.experiment("OAuth", function() {
     });
   });
 
-  lab.test("GET /user invalid auth", function(done) {
+  lab.test("GET /user returns 401 with malformed Authorization header", function(done) {
     ls.start(function(error) {
       var accessTokenRequest = {
         method: "GET",
@@ -653,7 +653,7 @@ lab.experiment("OAuth", function() {
     });
   });
 
-  lab.test("GET /user invalid auth", function(done) {
+  lab.test("GET /user returns 401 with malformed Authorization header", function(done) {
     ls.start(function(error) {
       var accessTokenRequest = {
         method: "GET",
@@ -670,7 +670,7 @@ lab.experiment("OAuth", function() {
     });
   });
 
-  lab.test("GET /user invalid auth", function(done) {
+  lab.test("GET /user returns 401 with invalid access token", function(done) {
     ls.start(function(error) {
       var accessTokenRequest = {
         method: "GET",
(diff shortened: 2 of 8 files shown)


toolness:
LOL, i love the commit messages. :+1:

cadecairos:
R?

test/web.js
```diff
@@ -616,4 +616,91 @@ lab.experiment("OAuth", function() {
...
+      };
+
+      s.inject(accessTokenRequest, function(response) {
+        Code.expect(response.statusCode).to.equal(401);
+        done();
+      });
+    });
+  });
+
+  lab.test("GET /user invalid auth", function(done) {
```
toolness: I like test names w/ descriptive titles b/c they end up being useful as documentation and they make it easier to figure out what test is failing... While I'm not familiar w/ your test framework, it seems like the above 4 tests are all called "GET /user invalid auth". Rather than that I might consider calling them e.g.:

* GET /user returns 401 when authorization header is malformed
* GET /user returns 401 when authorization header has no token
* GET /user returns 401 when authorization header has invalid token
* GET /user returns 401 when authorization header has expired token

That will make it easier for readers to understand what it is the test is trying to verify, and if the tests ever fail, it'll be easier for folks to understand exactly what is amiss from an "end-user" (or in this case an oauth2 client) perspective.

toolness: Obviously that's just a suggestion though, and probably shouldn't be a blocker to getting this merged!

toolness:
Hmm, so I have checked out this PR and am trying to manually integrate it w/ the teach API. But I'm finding that `GET /user`, when given the proper token, is returning the following:

```json
{"statusCode":404,"error":"Not Found","message":"LoginAPI error"}
```

I am pretty sure communication w/ the LoginAPI in general is working OK because I was able to log in... And if I change my token to something invalid I do get a 401, so I think I am hitting the right endpoint... Any idea what could be wrong?

web/server.js
```diff
@@ -291,6 +292,75 @@ module.exports = function(options) {
...
+                }
+
+                reply(token);
+              });
+            }
+          },
+          {
+            assign: 'user',
+            method: function(request, reply) {
+              account.getUser(request.pre.token.user_id, function(err, json) {
```
toolness: I *think* that `request.pre.token.user_id` is `undefined` here because I added some logging to `getUser` and it thinks its first argument is undefined. Which means that it's ultimately hitting `https://login.webmaker.org/user/username/undefined`, which would explain the weird [404 I'm getting](https://github.com/mozilla/id.webmaker.org/pull/128#issuecomment-90271655).

toolness: Um... adding further logging into the app, it seems like `token.user_id` is set to `undefined` fairly early on, at least as early as `OAuthDB.prototype.generateAuthCode`, which gets passed `undefined` for `userId`...

That function is being called by the handler for `GET /login/oauth/authorize`, which is passing it `request.pre.user.username`, but `request.pre.user` has no such key... In fact, it just looks like this when  logged to the console: `{ set: [Function], clear: [Function], ttl: [Function] }`. So that is weird.

lib/scopes.js
```diff
@@ -0,0 +1,29 @@
+var SCOPES = {
+  user: [
```
jbuck: :+1: This is a really easy way to see what scopes get you what data.

Is there any additional data we should add here? https://api.github.com/users/jbuck as inspiration

web/server.js
cadecairos: Turned out this was because the session data lives on `request.auth.credentials`, not `request.auth.session`

lib/scopes.js
cadecairos: So, What we currently have available from login.wmo is:

``` json
{
    "avatar": "https://secure.gravatar.com/avatar/5410ec660422c8ea12d4a28ef794a23a?d=https%3A%2F%2Fstuff.webmaker.org%2Favatars%2Fwebmaker-avatar-200x200.png",
    "email": "chris@chrisdecairos.ca",
    "emailHash": "5410ec660422c8ea12d4a28ef794a23a",
    "id": 1,
    "isAdmin": false,
    "isMentor": false,
    "isSuperMentor": false,
    "prefLocale": "en-US",
    "sendEventCreationEmails": true,
    "sendCoorganizerNotificationEmails": true,
    "sendMentorRequestEmails": true,
    "username": "cade"
  }
```
how about id, isAdmin, isMentor, isSuperMentor, & prefLocale,

jbuck: lets add `avatar` and `id` for now... I'd actually love to ditch those `is*` booleans, and I'm not sure how much sense `prefLocale` makes right now. We can always add it later

jbuck:
@cadecairos I think @toolness comment about the test names is a good fix, but this is good to go otherwise. r+ :+1:

PR #130, opened 2015-04-06
Validation fixes

alicoding:


alicoding:
@Pomax R?

templates/lib/validatorset.js
```diff
@@ -2,7 +2,7 @@ var Joi = require('joi');
 
 var fieldValidators = {
   username:       Joi.string().alphanum().min(1).required().label('Username'),
-  password:       Joi.string().min(8).required().label('Password'),
+  password:       Joi.string().regex(/^\S{8,16}$/).label('Password'),
```
Pomax: let's not put a max length restriction on passwords.

alicoding:
@Pomax I have updated the max_length to 128 now.

templates/components/form/form.jsx
```diff
@@ -66,9 +66,9 @@ var Form = React.createClass({
       input = (<span>{input}<span/></span>);
     }
 
-    var errorTooltip = (
-      <span className="warning">{value.errorMessage}</span>
-    );
+    var errorTooltip = (<span className="warning">{value.customError ? value.customError :
```
Pomax: nested ternary = let's not use nested ternary.
```
var errorTooltip = value.customError;
if (!errorTooltip) {
  errorTooltip = (id === 'password' ? 'Invalid password' : this.getValidationMessages(id)[0]);
}
errorTooltip = <span className="warning">{errorTooltip}<span>;
...
```
for instance? Also, those `==` should have been lint-caught... lastly, can we wrap this hardcoded string in a way that makes it easier to localise later?

templates/lib/validatorset.js
```diff
@@ -2,7 +2,7 @@ var Joi = require('joi');
 
 var fieldValidators = {
   username:       Joi.string().alphanum().min(1).required().label('Username'),
-  password:       Joi.string().min(8).required().label('Password'),
+  password:       Joi.string().regex(/^\S{8,128}$/).label('Password'),
```
Pomax: for future-proofing, we probably want `/.../u` here. Mostly because 128 single byte characters is not the same as 128 multibyte characters

templates/pages/migrate.jsx
```diff
@@ -40,31 +40,34 @@ var UserMigration = React.createClass({
...
   },
   handleLogin: function(error, data) {
-    console.log("inside App we see:", error, data);
-    if(!error) {
-      this.setState({
-        login: false,
-        setKey: true
-      });
+    if(error) {
+      console.log("inside App we see:", error, data);
```
Pomax: `return console.error(....)`?

templates/pages/migrate.jsx
```diff
@@ -40,31 +40,34 @@ var UserMigration = React.createClass({
...
   },
   handleSetKey: function(error, data) {
-    console.log("inside App we see:", error, data);
-    if(!error) {
-      this.setState({
-        setKey: false,
-        setPass: true
-      });
+    if(error) {
+      console.log("inside App we see:", error, data);
```
Pomax: see note above

templates/pages/migrate.jsx
```diff
@@ -40,31 +40,34 @@ var UserMigration = React.createClass({
...
   },
   handleResetPassword: function(error, data) {
-    console.log("inside App we see:", error, data);
-    if(!error) {
-      this.setState({
-        setPass: false,
-        success: true
-      });
+    if(error) {
+      console.log("inside App we see:", error, data);
```
Pomax: see note above

templates/components/set-password-migration-form.jsx
```diff
@@ -7,7 +7,7 @@ var fields = [
   {
     'password': {
       'placeholder': 'Enter your new password',
-      'type': 'text',
+      'type': 'password',
       'validator': 'username',
       'errorMessage': 'Invalid username'
```
Pomax: are we keeping error messages here, even though we remove them in login.jsx?

alicoding: Nice catch!

templates/components/form/form.jsx
```diff
@@ -49,6 +49,8 @@ var Form = React.createClass({
     var value = this.props.fields[i][id];
     this.passChecked = value.checked;
     this.beforeLabel = value.label === undefined ? true : value.label;
+    var invalidPWDSRT = 'Invalid password';
```
Pomax: let's give this a real name ;)

templates/lib/validatorset.js
```diff
@@ -2,7 +2,7 @@ var Joi = require('joi');
 
 var fieldValidators = {
   username:       Joi.string().alphanum().min(1).required().label('Username'),
-  password:       Joi.string().min(8).required().label('Password'),
+  password:       Joi.string().regex(/^\S{8,128}$/u).label('Password'),
```
Pomax: so, my bad, /u is not supported in every browser =_=

PR #135, opened 2015-04-07
Add Optimizely tag with env var to switch on / off

adamlofting:
Part of #77


alicoding:
r+

One question... Is there another way to load that script tag? I'm not really convince about the way we do it like that though I don't have enough knowledge as if there is a better way at all. I will research and if I find anything better will make a PR for you to look at, but for now lets land this.

adamlofting:
Yes, if we can find a better way to load the script tag please do. I couldn't find a very 'React' way to append it to the head conditionally.

PR #143, opened 2015-04-07
Add opens sans font

### LINKED ISSUE #141: Signup page: font not consistent across browsers
The font on signup page is different on Firefox (Helvetica) than it is on Chrome (Open Sans).

alicoding:
@secretrobotron R?
fixes #141


secretrobotron:
r+

PR #158, opened 2015-04-07
adds a simple way to quickly get up and running for local development

Pomax:


web/index.js
```diff
@@ -1,3 +1,7 @@
+var Habitat = require('habitat');
+Habitat.load();
+var habitat = new Habitat();
```
jbuck: if we're still relying on `process.env` then you don't need this line; `Habitat.load()` does the "load-file-into-process.env" bit

sample.env
```diff
@@ -0,0 +1,8 @@
+export HOST=0.0.0.0
+export PORT=1234
+export LOGINAPI=https://user:password@login.webmaker.org
+export COOKIE_SECRET=topSecretPasswordForEncryptingCookies
+export SECURE_COOKIES=true
```
jbuck: This should be defaulted to false, since the local server is probably running on http

sample.env
```diff
@@ -0,0 +1,8 @@
...
+export URI=https://id.webmaker.org
```
jbuck: This should be defaulted to http://localhost:1234

sample.env
```diff
@@ -0,0 +1,8 @@
+export HOST=0.0.0.0
+export PORT=1234
+export LOGINAPI=https://user:password@login.webmaker.org
```
jbuck: http://testuser:password@localhost:3000

jbuck:
r+, just change the defaults to match what the local dev defaults are

PR #161, opened 2015-04-07
 Buttons should be aligned with forms #153

### LINKED ISSUE #153: Buttons not aligned with forms
Buttons should be right-aligned with forms on all pages.

alicoding:
fixes #153


alicoding:
@secretrobotron R?

secretrobotron:
@alicoding the button on the signup page needs to be full-width and on top of the ToS notice.

With that, r+.

PR #167, opened 2015-04-08
Signup page: checkbox label misaligned #149

### LINKED ISSUE #149: Signup page: checkbox label misaligned
The label for the "Tell me about Mozilla" checkbox on the signup page is not vertically aligned with the checkbox.

@vazquez says you can just `float: left` to solve this problem.

alicoding:
fixes #149

### DIFF
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -65,7 +65,7 @@ var Form = React.createClass({
     );
 
     if (value.type === 'checkbox') {
-      input = (<span>{input}<span/></span>);
+      input = (<span className={value.className}>{input}<span/></span>);
     }
 
     var errorTooltip = value.customError;
--- a/templates/less/pages/signup.less
+++ b/templates/less/pages/signup.less
@@ -20,7 +20,16 @@
 
 
   .inputBox {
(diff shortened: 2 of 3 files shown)


alicoding:
@secretrobotron R?

templates/components/form/form.jsx
```diff
@@ -65,7 +65,7 @@ var Form = React.createClass({
     );
 
     if (value.type === 'checkbox') {
-      input = (<span>{input}<span/></span>);
+      input = (<span className={value.className}>{input}<span/></span>);
```
secretrobotron: Will this cause an error if `value.className` doesn't exist?

secretrobotron:
Just check out my comment above, and r+ :).

templates/components/form/form.jsx
alicoding: If it's undefined then it will just not add any class at all which shouldn't be a problem here since I've tested with other pages and we don't have any problem... It's also because we only use check box in sign up page too I guess...

secretrobotron:
Cool! Merge like you've never merged before! :rocket:

PR #169, opened 2015-04-08
TabIndex and accessibility for checkbox

### LINKED ISSUE #146: Signup page: fix tab order on form elements
The tab order on the signup page form elements should be go from Password :arrow_forward: Checkbox :arrow_forward: Button

alicoding:
fixes #146

### DIFF
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -60,8 +60,12 @@ var Form = React.createClass({
       <input type={value.type}
              id={id}
              ref={id+'Input'}
+             tabIndex={value.tabIndex}
              placeholder={value.placeholder}
              valueLink={this.linkState(id)}
+             role={value.type === 'checkbox' ? 'checkbox' : false}
+             area-checked={value.type === 'checkbox' ? 'false' : false}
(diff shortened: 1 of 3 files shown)


alicoding:
@secretrobotron R?

That CSS might look a bit hacky but that's how we can achieve accessibility and being fancy at the same time >.<

secretrobotron:
This works well. Couple of comments though:
- There's no way to tell if the checkbox is actually highlighted when you move from password to checkbox
- After the checkbox, shouldn't the next item in tab order be the button, instead of the ToS link?

alicoding:
I have fixed the checkbox with adding focus class to it.

After the checkbox I think it should be Tos before button since we want them to read it before signing up, no?

secretrobotron:
They *should* read it, but I think it's a bit counter-intuitive as is. I'd expect the button to be highlighted next because it's most important. If I want to read the ToS and Privacy policy, I can click on those.

alicoding:
@secretrobotron that would be something you see, but for a11y wouldn't that skip the text and make it not even exist for them because they're already on the button?

secretrobotron:
Ok, no worries about enter button default for now. :+1:

Issue #109, opened 2015-04-01
Create user migration email


secretrobotron:
- [x] Find a reference email
- [x] Create copy for it
- [x] File bugs to hook it up
- [ ] IMPLEMENT

secretrobotron:
Reference email: https://github.com/mozilla/webmaker-mailroom/blob/master/locale/en_US/login_request.json

secretrobotron:
Current login reset email looks like this:
![screenshot 2015-04-01 15 51 28](https://cloud.githubusercontent.com/assets/403231/6950182/0c315c0c-d887-11e4-9952-61e8f7735df0.png)

secretrobotron:
From Kevin:
> Webmaker is using a new log in system to ensure your experience is simple and safe. To get started, click here to set a password, or copy and paste this migration token into the text field:  xxx

I'd leave the message at the bottom too:

> If you did not try to sign in, you can ignore this message, or send an email to help@webmaker.org to ask for help.

secretrobotron:
@cadecairos can you list the patches you'll need to make for this (like you did on our call yesterday)?

cadecairos:
1. Login.wmo - needs a way to indicate that sawmill should render a migration token email VS regular token login email
2. webmaker-mailroom - migration email template, strings
3. sawmill - check for flag on token login emails to trigger a switch to the migration email template.

secretrobotron:
Subject line: "Set up your new Webmaker account password"

Remove the second button. Just make the first one say "Set Password".

secretrobotron:
@KevZawacki looking at, "copy and paste this migration token into the text field," it might make more sense as "copy and paste this migration token into the text field on [link to set password page]this page[/link]".

KevZawacki:
@secretrobotron +1, agreed

cadecairos:
bloker bugs in login, sawmill and webmaker-mailroom are all landed, I'm working on wiring everything together.

cadecairos:
fixed in #194

PR #194, opened 2015-04-08
Migrate all the users

cadecairos:


lib/account.js
```diff
@@ -162,6 +162,101 @@ module.exports = function(config) {
...
+      var migrateRequest = hyperquest({
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'x-ratelimit-ip': getIPAddress(request)
+        },
+        uri: loginAPI + '/api/v2/user/request'
+      });
+
+      migrateRequest.on('error', callback);
```
secretrobotron: Does this include timeouts?

lib/account.js
```diff
@@ -162,6 +162,101 @@ module.exports = function(config) {
...
+
+      passwordRequest.on('error', callback);
+
+      passwordRequest.on('response', function(message) {
+        parseMessage(message, function(err, json) {
+          callback(err, json);
+        });
+      });
+
+      passwordRequest.end(JSON.stringify({
```
secretrobotron: So, when this happens, is there something on `loginAPI` that checks whether or not we should really allow password-setting here?

cadecairos: allow/disallow in what context?

lib/account.js
cadecairos: https://www.npmjs.com/package/hyperquest#req-on-error-function-res

If I understand that correctly, if an http request times out, it will be passed via the error event.

lib/account.js
secretrobotron: Mmm I see that `request.auth.isAuthenticated` prevents it on this side in `server.js`. There's no way someone can construct a POST from cli to tell loginAPI to reset password though, right?

lib/account.js
secretrobotron: :+1:

lib/account.js
cadecairos: right, the actual loginAPI endpoint is protected with basic authentication.

secretrobotron: (｀_´)ゞ

templates/components/login-no-pass-form.jsx
```diff
@@ -18,8 +18,8 @@ var LoginNoPassword = React.createClass({
   render: function() {
     return (
       <div className="loginNoPass innerForm centerDiv">
-        <Form origin="Migration" ref="userform" fields={fields} validators={fieldsValidators} />
-        <button onClick={this.processFormData} className="btn btn-awsm">Log in</button>
+        <Form origin="Migration" ref="userform" fields={fields} validators={fieldsValidators} defaultUsername={this.props.username} />
```
alicoding: Just a reminder for you and me that we'll need `IconText` component here with the text @secretrobotron gave us yesterday.

secretrobotron: @alicoding did you write down what that text was?

alicoding: I think I lost that when we closed the vidyo, but I remembered that we need to change it after you've sent the first copy because of some reason... >.<

alicoding:
This should solve #53

templates/components/login-no-pass-form.jsx
secretrobotron: > Webmaker is using a new log in system using passwords to ensure your experience is simple and safe. To get started, verify your username below, and click the "Set Password" button.

secretrobotron:
Will fix #109.

web/server.js
```diff
@@ -373,6 +366,102 @@ module.exports = function(options) {
...
+          },
+          {
+            assign: 'password',
+            method: function(request, reply) {
+              var password = request.payload.password;
+              if ( !password ) {
+                return reply(Boom.badRequest('No password provided'));
+              }
+
+              var result = passTest.test(password, [request.pre.username]);
```
cadecairos: add email here

PR #198, opened 2015-04-10
Add validation user migration

alicoding:


templates/lib/validatorset.js
```diff
@@ -2,7 +2,7 @@ var Joi = require('joi');
 
 var fieldValidators = {
   username:       Joi.string().alphanum().min(1).required().label('Username'),
-  password:       Joi.string().regex(/^\S{8,128}$/).label('Password'),
+  password:       Joi.string().regex(/^\S{1,128}$/).label('Password'),
```
secretrobotron: Should probably leave this at 8, no?

alicoding: ah, right! I forgot to switch this back after those testings :P

PR #202, opened 2015-04-10
Bump version of react-ga

adamlofting:
r?

I fixed a bug in `react-ga` today which was stopping the 'labels' from being recorded in GA events.


alicoding:
r+

PR #207, opened 2015-04-13
minify and optimize app.bundle.js

### LINKED ISSUE #200: Perf: cut down on size
screenshot 2015-04-09 22 56 58

- why are there map files being downloaded? why are they part of dist?
- are we using an compiler/uglifier to shrink the size of `app.bundle.js`?

cadecairos:
Fixes #200 

running `npm run start:production` will produce an app.bundle.js of 529KB VS 1192KB


cadecairos:
R?

jbuck:
@cadecairos we need to think about how to get this to run on Heroku as well

cadecairos:
well how does heroku start up things? plain 'ol npm start?

cadecairos:
looks like I can update the postinstall script to be the prod build. unfortunately, that means npm install will be slowed down significantly.

cadecairos:
oh wait, I already did. rejoice! :fireworks:

jbuck:
It uses the command specified in the Procfile to run the app, but the
problem here is building the minified code:
https://devcenter.heroku.com/articles/nodejs-support#build-behavior

I'm not sure of the best way to handle this. If we naively build with
optimization every time, that'll suck for development.

package.json
```diff
@@ -5,20 +5,24 @@
...
-    "postinstall": "npm run build",
+    "postinstall": "node build",
     "pretest:browser": "ncp node_modules/mocha/mocha.js public/tests/mocha.js && ncp node_modules/mocha/mocha.css public/tests/mocha.css",
     "test:browser": "npm run build && mocha-phantomjs -s loadImages=false public/tests/index.html",
     "test:server": "lab -t 95",
     "watch:js": "npm run build:js -- --watch",
-    "watch:css": "autoless --source-map templates/less public"
+    "watch:js:production": "npm run build:js:production -- --watch",
+    "watch:css": "autoless --source-map templates/less public",
+    "test": "npm run test:server && npm run test:browser && npm run lint"
```
jbuck: Test should go back up to the top

package.json
```diff
@@ -5,20 +5,24 @@
   "main": "index.js",
   "scripts": {
     "start": "parallelshell \"node web/index.js\" \"npm run watch:css\" \"npm run watch:js\"",
-    "test": "npm run test:server && npm run test:browser && npm run lint",
+    "start:production": "parallelshell \"node web/index.js\" \"npm run watch:css\" \"npm run watch:js:production\"",
     "build": "npm run build:css && npm run build:js",
+    "build:production": "npm run build:css && npm run build:js:production",
     "build:css": "npm run watch:css -- --no-watch",
     "build:js": "webpack --config webpack.config.js --progress --profile --colors",
+    "build:js:production": "webpack --config webpack.config.js --progress --profile --colors  --optimize-minimize --optimize-dedupe",
```
jbuck: You could change this to `npm run build:js -- --optimize-minimize --optimize-dedupe`

jbuck: Actually, just move the additional arguments to the `build:production` script

package.json
```diff
@@ -5,20 +5,24 @@
   "main": "index.js",
   "scripts": {
     "start": "parallelshell \"node web/index.js\" \"npm run watch:css\" \"npm run watch:js\"",
-    "test": "npm run test:server && npm run test:browser && npm run lint",
+    "start:production": "parallelshell \"node web/index.js\" \"npm run watch:css\" \"npm run watch:js:production\"",
```
jbuck: I don't think we need a start:production task now, do we?

cadecairos:
@jbuck moved things around, made fixes. R?

jbuck:
@cadecairos r+, but I'd like to check with @Pomax that we did the node script correctly for Windows.

PR #221, opened 2015-04-14
Fix Incorrect console error after migrate flow

### LINKED ISSUE #220: Incorrect console error after migrate flow
This appears in the console at the end of the `/migrate` flow regardless of success:
```
Non 200 statusCode recieved while attemting migration undefined
```

My guess is this is the problem in `migrate.jsx`:
```
	      } else if (json.statusCode !== 200) {
```
where `json.statusCode` is probably undefined.

alicoding:
Fixes #220

### DIFF
--- a/templates/pages/migrate.jsx
+++ b/templates/pages/migrate.jsx
@@ -127,17 +127,19 @@ var UserMigration = React.createClass({
       }
       return response.json();
     }).then((json) => {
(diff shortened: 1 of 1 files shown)


alicoding:
@secretrobotron R?

secretrobotron:
r+

PR #224, opened 2015-04-15
CSRF Protection

### LINKED ISSUE #213: Implement CSRF
- [ ] find out how this will work with the react router
- [ ] should do csrf on per-request basis (rather than per-page-load)?
 - is this really more secure?
 - hash per request or secret per request?

cadecairos:
Fixes #213


cadecairos:
R? @jbuck and/or @simonwex

cadecairos:
@jbuck crumb does not provide an API for regenerating the crumb manually, AFAICT

simonwex:
@cadecairos I think an automated test here should be pretty easy to write and pretty important to catch regressions if we screw with headers in the future.

cadecairos:
@simonwex Added two tests for each POST request endpoint and a test to check that the crumb cookie is set when a GET request is received

web/server.js
```diff
@@ -45,6 +45,20 @@ module.exports = function(options) {
     });
   });
 
+  function skipCSRF(request, reply) {
+    return reply(true);
```
jbuck: Is this correct? From the docs:

```
'skip' - a function with the signature of function (request, reply) {}, which when provided, is called for every request. If the provided function returns true, validation and generation of crumb is skipped (defaults to false)
```

I think you just need to `return true` here?

jbuck: Yeah, looking at the code, this needs to be fixed up: https://github.com/hapijs/crumb/blob/master/lib/index.js#L62

jbuck:
@cadecairos r+

PR #228, opened 2015-04-15
Fix #222 - Email validation doesn't check for uniqueness

### LINKED ISSUE #222: Email validation doesn't check for uniqueness
Following from #178, `/signup` should also check to see if the email you use is unique (because of current loginapi server restriction).

alicoding:
fixes #222

### DIFF
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -53,12 +53,13 @@ var Form = React.createClass({
       key: '',
       errorMessage: {},
       valid_username: false,
+      valid_email: false,
       passwordError: false
     };
   },
   setFormState: function(data) {
     this.setState({
-      valid_username: true,
+      ['valid_'+data.field]: true,
       errorMessage: {[data.field]: null}
     });
   },
@@ -73,10 +74,21 @@ var Form = React.createClass({
   dirty: function(id, origin) {
     return (err, valid) => {
       if(err) {
+        if(id === 'email') {
+          this.setState({
+            ['valid_'+id]: false,
+            errorMessage: {
+              [id]: 'Please use a valid email address.'
+            }
+          });
+        }
         ga.event({category: origin, action: 'Validation Error', label: 'Error on ' + id + ' field.'});
       }
+      if(!err && id === 'email') {
+        this.setFormState({field: 'email'})
+      }
       this.state.passwordError = null;
-      this.handleBlur(id, this.state.username)
+      this.handleBlur(id, this.state[id])
     }
   },
   handleBlur: function(fieldName, value) {
@@ -103,7 +115,7 @@ var Form = React.createClass({
     var passwordError = this.state.passwordError;
 
     if(id === 'password' && !this.isValid(id) && !passwordError) {
-      passwordError = 'Invalid Password';
+      passwordError = 'Invalid password.';
     }
 
     var isValid = !this.state.errorMessage[id] && this.isValid(id) && !passwordError;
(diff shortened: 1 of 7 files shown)


alicoding:
@secretrobotron R?

alicoding:
or @cadecairos R?

secretrobotron:
Gotta fix up those tests, it seems.

alicoding:
@secretrobotron done! :D

templates/components/form/form.jsx
```diff
@@ -53,12 +53,13 @@ var Form = React.createClass({
...
       errorMessage: {},
       valid_username: false,
+      valid_email: false,
       passwordError: false
     };
   },
   setFormState: function(data) {
     this.setState({
-      valid_username: true,
+      ['valid_'+data.field]: true,
```
cadecairos: nice. can you add some whitespace here?

templates/components/form/form.jsx
```diff
@@ -73,10 +74,21 @@ var Form = React.createClass({
   dirty: function(id, origin) {
     return (err, valid) => {
       if(err) {
+        if(id === 'email') {
+          this.setState({
+            ['valid_'+id]: false,
```
cadecairos: if we know id is email, why not just key it as 'valid_email'

templates/components/form/form.jsx
```diff
@@ -73,10 +74,21 @@ var Form = React.createClass({
...
+            errorMessage: {
+              [id]: 'Please use a valid email address.'
```
cadecairos: same deal here

secretrobotron:
Looks good. One problem I see:

1. Go to /signup.
2. Type in username that already exists.
3. Tab to email box (error should appear in username box).
4. Type in email that already exists.
5. Tab to password box (error should appear in email box).

The error for username will disappear:

screenshot 2015-04-16 09 12 21

Then, I can end up with a form that looks like this that won't submit, but there are no errors :(:
screenshot 2015-04-16 09 14 42

test/web.js
```diff
@@ -73,28 +73,30 @@ lab.experiment("OAuth", function() {
       };
 
       s.inject(request, function(response) {
-        Code.expect(response.statusCode).to.equal(500);
+        Code.expect(response.statusCode).to.equal(400);
+        Code.expect(response.result.message).to.equal("Password not strong enough.");
```
cadecairos: give this test case a valid password, change it back to expecting a 500

test/web.js
```diff
@@ -73,28 +73,30 @@ lab.experiment("OAuth", function() {
...
       Code.expect(error).to.be.undefined();
       var request = {
         method: "POST",
         url: "/create-user",
         payload: {
-          email: "webmaker@example.com",
+          email: "webmaker",
           username: "notgonnawork",
-          password: "CantGuessThis",
+          password: "CantGuessThis1234",
```
cadecairos: keep this valid password, fix the email param and title back to what they were, and then undo the changes to the assertions.

web/server.js
```diff
@@ -260,6 +260,33 @@ module.exports = function(options) {
...
+                var err = Boom.badRequest('Password not strong enough.', result);
+                err.output.payload.details = err.data;
+                return reply(err);
+              }
+
+              reply(password);
+            }
+          }
+        ],
         validate: {
```
cadecairos: move this above the pre section.

web/server.js
```diff
@@ -260,6 +260,33 @@ module.exports = function(options) {
...
+            assign: 'username',
+            method: function(request, reply) {
+              reply(request.payload.username);
+            }
+          },
+          {
+            assign: 'password',
+            method: function(request, reply) {
+              var password = request.payload.password;
+              if ( !password ) {
```
cadecairos: we don't need this check, since we're guaranteed it'll be here by the payload validation config

PR #240, opened 2015-04-17
Fix issue197 - change form to be an actual HTML form

### LINKED ISSUE #197: Firefox password manager doesn't work with login form
I've noticed that Firefox doesn't ask me if I want it to remember my username/password when I submit the login form, which is a bummer. Haven't tried it on other browsers though.

alicoding:
Fixes #197

### DIFF
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -160,7 +160,7 @@ var Form = React.createClass({
   },
   render: function() {
      var fields = Object.keys(this.props.fields).map(this.buildFormElement);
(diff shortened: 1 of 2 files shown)


alicoding:
@secretrobotron r?

@toolness I'm not sure if this is really going to fix the issue since I don't use Firefox Password Manager and not sure how to trigger that, but for my lastpass at least it's showing me the field properly now where as before it wasn't.

alicoding:
@cadecairos r? too if you can

cadecairos:
R+

PR #248, opened 2015-04-17
Use // for font's url

alicoding:


alicoding:
@cadecairos R?

templates/less/styles.less
```diff
@@ -3,7 +3,7 @@
 @import "../../node_modules/font-awesome/less/font-awesome";
 @fa-font-path: "/assets/font-awesome/fonts";
 // Open Sans font
-@import url(http://fonts.googleapis.com/css?family=Open+Sans:400,300,400italic,300italic,700);
+@import url(//fonts.googleapis.com/css?family=Open+Sans:400,300,400italic,300italic,700);
```
jbuck: You should force this to be https instead: http://www.paulirish.com/2010/the-protocol-relative-url/

alicoding: Good point!

alicoding:
@jbuck r?

PR #249, opened 2015-04-17
no csrf for access_token

cadecairos:


jbuck:
r-, you should use this approach instead: https://github.com/hapijs/crumb/issues/32#issuecomment-61173586

alicoding:
r+ if travis is happy with your hapi and I will be happy!

PR #258, opened 2015-04-20
Make sure no background from signup page cross over to login

### LINKED ISSUE #261: Background image on login page
There shouldn't be any background image on /login but when you visit /signup then switch to /login you will see background image.

alicoding:
fixes #261


secretrobotron:
@alicoding does this fix a bug with a specific number?

PR #267, opened 2015-04-20
Fix issue #253

### LINKED ISSUE #253: Pressing enter doesn't submit forms in Safari
On /login, enter button doesn't submit forms. I suspect the same problem exists on all pages.

alicoding:
fixes #253

### DIFF
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -38,10 +38,14 @@ var Form = React.createClass({
   componentDidMount: function() {
     WebmakerActions.addListener('FORM_ERROR', this.formError);
     WebmakerActions.addListener('FORM_VALID', this.setFormState);
+    window.addEventListener("keyup", this.processFormData);
+    window.addEventListener("keydown", this.onKeyDownTriggered);
   },
   componentWillUnmount: function() {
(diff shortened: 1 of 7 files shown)


alicoding:
@secretrobotron r?

templates/components/form/form.jsx
```diff
@@ -224,11 +231,14 @@ var Form = React.createClass({
...
    * "owner" components call form.processFormData on us
    */
-  processFormData: function(e) {
-    e.preventDefault();
-    this.validate((error, data) => {
-      WebmakerActions.onFormValidation({err: error, user: !!error ? false : JSON.parse(JSON.stringify(this.state))});
-    });
+  processFormData: function(evt, clicked) {
+    evt.preventDefault();
+    let keyCode = evt.keyCode;
```
secretrobotron: Don't think you need to cache this var if it's just used once below. Am I missing something you're doing with it here?

alicoding: I'm only using this in the same scope which should be fine?

alicoding: ignore my message here. I'm a bit blur LOL

PR #271, opened 2015-04-20
Fix issue#259 - Default username after reset password

### LINKED ISSUE #259: Blank /login screen shown after password reset
After resetting password, user is shown blank login screen. Can we at least provide the username so the user doesn't have to type it again? (Would be ideal to just log them in. @cadecairos ?)

alicoding:
Fixes #259


alicoding:
@secretrobotron r?

secretrobotron:
:+1:

PR #278, opened 2015-04-21
Fix #262 - allow hyphen in the username field

### LINKED ISSUE #262: Users can't migrate with usernames with non-alpha-numeric characters
screenshot 2015-04-20 13 51 00

@alicoding @cadecairos @thisandagain might be a problem with only accepting alpha-numeric characters. At the very least, people can't migrate. What should they do?

alicoding:
fixes #262

### DIFF
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -78,6 +78,11 @@ var Form = React.createClass({
   dirty: function(id, origin) {
     return (err, valid) => {
       if(err) {
+        if(id === 'username' && this.state[id]) {
+          this.formError({field: 'username', message: 'Must be 1-20 characters long and use only "-" and alphanumeric symbols.'});
+          // preventing any message to override this error message.
+          return;
+        }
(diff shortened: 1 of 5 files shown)


alicoding:
@secretrobotron R?

templates/components/form/form.jsx
```diff
@@ -78,6 +78,10 @@ var Form = React.createClass({
   dirty: function(id, origin) {
     return (err, valid) => {
       if(err) {
+        if(id === 'username' && this.state[id]) {
+          this.formError({field: 'username', message: 'Username must be between 1-20 characters, and may only include \"-\" and alphanumeric symbols.'});
```
secretrobotron: `Must be between 1-20 characters, and use only \"-\" and alphanumeric symbols.`

secretrobotron: `Must be 1-20 characters long and use only "-" and alphanumeric symbols.`

secretrobotron:
r+

PR #279, opened 2015-04-21
Content Security Policy

### LINKED ISSUE #270: Add CSP

cadecairos:
fixes #270

### DIFF
--- a/web/server.js
+++ b/web/server.js
@@ -32,7 +32,41 @@ module.exports = function(options) {
     port: options.port
   });
 
-  server.register(require('hapi-auth-cookie'), function(err) {
+  server.register([
+    require('hapi-auth-cookie'),
+    require('scooter'),
+    {
+      register: require('blankie'),
+      options: {
+        defaultSrc: [
+          '\'self\''
+        ],
+        styleSrc: [
+          '\'self\'',
+          'https://fonts.googleapis.com'
+        ],
+        imgSrc: [
+          '\'self\'',
+          'data:',
+          'https://www.google-analytics.com',
+          'http://www.google-analytics.com'
+        ],
+        frameSrc: [
+          '\'none\''
+        ],
+        scriptSrc: [
+          '\'self\'',
+          '\'unsafe-eval\'',
+          'https://www.google-analytics.com',
+          'http://www.google-analytics.com'
+        ],
+        fontSrc: [
+        '\'self\'',
+          'https://fonts.gstatic.com'
+        ]
+      }
+    }
+  ], function(err) {
     // MAYDAY, MAYDAY, MAYDAY!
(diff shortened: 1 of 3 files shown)


web/server.js
```diff
@@ -32,7 +32,41 @@ module.exports = function(options) {
...
+          'data:',
+          'https://www.google-analytics.com',
+          'http://www.google-analytics.com'
+        ],
+        childSrc: [
+          '\'none\''
+        ],
+        scriptSrc: [
+          '\'self\'',
+          '\'unsafe-eval\'',
```
jbuck: :weary: 

What requires this?

cadecairos: Joi D:

web/server.js
```diff
@@ -32,7 +32,41 @@ module.exports = function(options) {
...
+          '\'self\'',
+          'https://fonts.googleapis.com'
+        ],
+        imgSrc: [
+          '\'self\'',
+          'data:',
+          'https://www.google-analytics.com',
+          'http://www.google-analytics.com'
+        ],
+        childSrc: [
```
jbuck: Hm, any reason to call this out as `'non'`?

web/server.js
```diff
@@ -32,7 +32,41 @@ module.exports = function(options) {
...
 
-  server.register(require('hapi-auth-cookie'), function(err) {
+  server.register([
+    require('hapi-auth-cookie'),
+    require('scooter'),
+    {
+      register: require('blankie'),
+      options: {
+        defaultSrc: [
+          '\'self\''
```
jbuck: Could we use double quotes here, or is that a code style error?

web/server.js
```diff
@@ -32,7 +32,41 @@ module.exports = function(options) {
...
+          '\'none\''
+        ],
+        scriptSrc: [
+          '\'self\'',
+          '\'unsafe-eval\'',
+          'https://www.google-analytics.com',
+          'http://www.google-analytics.com'
+        ],
+        fontSrc: [
+        '\'self\'',
```
jbuck: indent is off here

web/server.js
cadecairos: well, according to MDN it'll fall back on default-src, which would be 'self'.

will we ever even want iframes?

web/server.js
cadecairos: JSCS FTL

web/server.js
cadecairos: look again

web/server.js
jbuck: fml

web/server.js
jbuck: Lets set default-src to none perhaps?

PR #293, opened 2015-05-13
first pass at password grant type

cadecairos:


web/server.js
```diff
@@ -209,13 +232,25 @@ module.exports = function(options) {
...
             method: function(request, reply) {
               oauthDb.getClient(request.payload.client_id, function(err, client) {
                 if ( err ) {
                   return reply(err);
                 }
-                if ( client.client_secret !== request.payload.client_secret ) {
+                if (
+                  client.allowed_grants.indexOf(request.pre.grant_type) === -1 ||
+                  (
+                    request.pre.grant_type === 'authorization_code' &&
```
jbuck: Maybe we should move this logic into the route handler, as opposed to doing it in the pre method

web/server.js
```diff
@@ -225,6 +260,19 @@ module.exports = function(options) {
...
             method: function(request, reply) {
+              if ( request.pre.grant_type === 'password' ) {
+                return account.verifyPassword(request, function(err, json) {
+                  if ( err ) {
+                    return reply(err);
+                  }
+
+                  reply({
+                    user_id: json.user.username,
+                    scopes: request.payload.scopes
```
jbuck: This looks like a red flag to me, blindly allowing any scopes. Is there any way we can limit this? Maybe take the scopes from the oauth client whitelist? I can't think of any other option.

cadecairos: we could ask for clients to declare scopes in advance for the password grant. I just followed what the spec said to do.

jbuck: I know you did, it's just that the spec is really light on details that make the handshake secure :)

web/server.js
cadecairos: That would require removing all pre methods below into the handler. are we okay with that?

cadecairos:
@jbuck I added some more commits, take a look.

web/server.js
jbuck: On second thought, if you have the username and password you could set whatever scope you wanted to anyways. Nevermind!

web/server.js
jbuck: I think we should do it, but lets do it as a follow-up.

docs/oauth.md
```diff
@@ -202,6 +202,62 @@ To log out of id.webmaker.org, use the /logout route
 After a successfully clearing the session, you'll be redirected to the redirect_uri of the supplied client_id, with a `logout=true` in the search portion of the redirect URL
 
 e.g `https://example.org/oauth_redirect?logout=true`
+
+## Native Application Flow
+
+Native applications can use the password grant type to authenticate Webmaker users, without compromising their client_secret.
```
jbuck: Should we do something like remove CORS headers on the `/login/oauth/access_token` route to ensure this?

jbuck:
@cadecairos r+

PR #298, opened 2015-05-15
Implement postgreSQL as a datastore for clients, auth codes & access …

### LINKED ISSUE #296: Token Storage DB
A postgresql db for storing tokens.

tokens are going to be salted with an in-memory secret known only to the process, hashed, and then stored. This way an attacker must do twice the work to compromise accounts, and we can invalidate all tokens in the wild by simply changing the secret.

cadecairos:
…tokens

Fixes #296


lib/oauth-db.js
```diff
@@ -1,35 +1,55 @@
...
 var Hoek = require('hoek');
+var pg = require('pg');
+var queries = require('./queries');
+var connString = process.env.POSTGRE_CONNECTION_STRING;
 
-var OAuthDB = function OAuthDB(clients, authorizations, tokens) {
-  this.clients = {};
-  this.authCodes = {};
-  this.accessTokens = {};
+var secret = process.env.SERVER_SECRET;
```
jbuck: Could you make this a more descriptive environment variable (TOKEN_SALT?)

jbuck: Also, is there a recommendation for how much entropy this should have to be considered secure?

lib/queries.js
```diff
@@ -0,0 +1,18 @@
...
+    accessToken: "SELECT access_token, client_id, user_id, scopes, created_at FROM " +
+      "access_tokens WHERE access_token = $1;"
+  },
+  create: {
+    client: "INSERT INTO clients VALUES($1, $2, $3, $4);",
+    authCode: "INSERT INTO auth_codes VALUES($1, $2, $3, $4, $5);",
+    accessToken: "INSERT INTO access_tokens VALUES($1, $2, $3, $4);"
+  },
+  update: {
+    authCode: "UPDATE auth_codes SET used = TRUE WHERE auth_code = $1;"
```
jbuck: Maybe we should delete the auth code instead of updating a bool? Less storage in the db, and I actually think you're not checking the "used" parameter in the "verifyAuthCode" code

lib/oauth-db.js
```diff
@@ -38,65 +58,111 @@ OAuthDB.prototype.generateAuthCode = function(clientId, userId, scopes, expiresA
...
+      });
+    }
+
+    if ( code.expires_at <= Date.now() ) {
+      return process.nextTick(function() {
+        callback(Boom.forbidden('auth code expired'));
+      });
+    }
+
+    process.nextTick(function() {
```
jbuck: You should delete the authCode or update the used column after fetching it w/o error

cadecairos: Meant to do that, the sql is already prepared

lib/oauth-db.js
cadecairos: I'd have to research that, but i think 256 bits as hex should be finee

cadecairos:
@jbuck Fixed everything up.

lib/oauth-db.js
```diff
@@ -1,35 +1,55 @@
...
+Hoek.assert(salt, 'You must define a TOKEN_SALT for salting auth codes and access tokens');
 
-  Object.keys(authorizations).forEach(function(auth) {
-    this.authCodes[auth] = authorizations[auth];
-  }, this);
-
-  tokens.forEach(function(token) {
-    this.accessTokens[token.access_token] = token;
-  }, this);
+var OAuthDB = function OAuthDB(Adapter) {
```
jbuck: Is this Adapter parameter being used for something? Maybe just needs to be removed?

cadecairos: cruft

lib/oauth-db.js
```diff
@@ -38,65 +58,117 @@ OAuthDB.prototype.generateAuthCode = function(clientId, userId, scopes, expiresA
...
 
-    this.authCodes[authCode] = {
-      client_id: clientId,
-      user_id: userId,
-      scopes: scopes,
-      expires_at: expiresAt
-    };
-
-    callback(null, authCode);
+    var saltedCode = crypto.createHash('sha256')
```
jbuck: I *think* we want to switch this to an [hmac generator](https://nodejs.org/api/crypto.html#crypto_crypto_createhmac_algorithm_key). Basically you input the salt as the key in the hmac instead. I want to find a concrete resource for *why* we should do this, but this comment on HN seems to explain the difference the best, and why we should use HMAC: https://news.ycombinator.com/item?id=1998247

lib/oauth-db.js
```diff
@@ -38,65 +58,117 @@ OAuthDB.prototype.generateAuthCode = function(clientId, userId, scopes, expiresA
...
+      clientId,
+      userId,
+      JSON.stringify(scopes),
+      expiresAt
+    ], function(err) {
+      if ( err ) {
+        return callback(err);
+      }
+
+      process.nextTick(function() {
```
jbuck: You can remove these `process.nextTick` wrappers for those queries. I originally did it that way since the in-memory DB i created would've been synchronous otherwise. Since doing the postgres query will always be asynchronous, no need for this!

scripts/create-tables.js
```diff
@@ -0,0 +1,15 @@
+var spawn = require('cross-spawn');
+
+var create = spawn('psql', ['-d','webmaker_oauth_test', '-f', 'scripts/create-tables.sql']);
+
+create.stdout.on('data', function(data) {
```
jbuck: I think you could replace these pipes with the `stdio` option: https://nodejs.org/api/child_process.html#child_process_options_stdio (specifically the stdio: "inherit" option)

jbuck:
@cadecairos this is looking good, just need a few small changes made.

cadecairos:
@jbuck plz review asap

jbuck:
@cadecairos r+

PR #308, opened 2015-05-21
Attach token client_id to /user reponse object

cadecairos:


jbuck:
@cadecairos I think we ended up deciding that we didn't need this PR. Let me know if I'm wrong!

PR #315, opened 2015-06-02
disable crumb and enable cors on POST /create-user

### LINKED ISSUE #314: Remove CSRF for create-user routes
Needed for native flow

cadecairos:
Fixes #314

### DIFF
--- a/test/web.js
+++ b/test/web.js
@@ -127,28 +127,6 @@ lab.experiment("OAuth", function() {
     });
   });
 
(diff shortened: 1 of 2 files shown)


cadecairos:
@jbuck see any issues with doing this?

k88hudson:
so, turns out we can actually get around CORS android-side -- so the only thing that's left is the CSRF stuff.

jbuck:
Lets make @k88hudson's suggested change so we don't add CORS headers, and land it

PR #319, opened 2015-06-11
I had no idea what I'm doing, but now I do

cadecairos:


cadecairos:
filed #320 to fix up the front end

thisandagain:
Nice. Working locally. R+

I think I'd like a second set of eyes on this though ... @k88hudson R?

k88hudson:
Looks good to me R+

PR #321, opened 2015-06-16
Pass a more meaningful error from login server

cadecairos:
mozilla/webmaker-android#2216

### DIFF
--- a/web/server.js
+++ b/web/server.js
@@ -434,6 +434,11 @@ module.exports = function(options) {
             return reply(err);
           }
           if ( json.login_error ) {
+            if ( json.login_error.indexOf('Users.username') !== -1 ) {
+              return reply(Boom.badRequest('That username is taken'));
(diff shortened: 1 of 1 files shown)


k88hudson:
Any we could change he error message from 

```
Error: SQLITE_CONSTRAINT: UNIQUE constraint failed: Users.username
``

to something like `That username is taken`?

cadecairos:
@k88hudson  We can do that here, or in  mozilla/webmaker-android#2216, Where would you prefer?

k88hudson:
@cadecairos I think it would be better to do on id.webmaker.org since error messages might change here/get out of sync in the app etc.

cadecairos:
okay, I'll check for each type of constraint error here.

k88hudson:
Thanks!

k88hudson:
R+ Awesome, thanks so much for doing this!

PR #323, opened 2015-06-18
redirect to android app if android=true on queryObj

cadecairos:


k88hudson:
R+, i'll do the follow-up patch

PR #334, opened 2015-07-08
Fix reset

### LINKED ISSUE #317: Can't recover password if I forgot my username
Lost password form requires a username. Can we set this to accept either username or email for ID? If not, email might be the better option since recovery will happen using it. It must be known.

cadecairos:
Fixes #317


alicoding:
r+

xmatthewx:
@cadecairos - I couldn't break it. Just found one small thing below.  

@flukeout - you're better at breaking things. try the pwd reset via https://teach.mofostaging.net/.

#### this is great
screenshot 2015-07-08 18 42 23

#### this could be better

"...with that email address"
or "...with that username or email address"

screenshot 2015-07-08 18 37 52

xmatthewx:
cc @secretrobotron

PR #342, opened 2015-07-28
Make lang an optional attribute for create user

jbuck:
lang must be an optional attribute for compatibility with older clients.
If we default to to en-US we get the desired behaviour: en-US for old
clients and new clients can set their preferred locale.

Fix https://github.com/mozilla/webmaker-core/issues/423 .


ashleygwilliams:
r+ looks good! :+1:

Issue #356, opened 2015-08-07
Enable Single Page App login Flow


cadecairos:
as per the details in:

http://aaronparecki.com/articles/2012/07/29/1/oauth2-simplified#browser-based-apps

we will add the option to pass response_type as "token" and for authorized client_ids, we'll return an api token in a url fragment to the callback url `https://example.org/singlepageapp#token=API_TOKEN` which should be the url of the single-page-app

cc/ @ashleygwilliams @jbuck

jbuck:
I can dig it!

cadecairos:
the code handling `/authorize` is pretty horrible...

PR #357, opened 2015-08-07
Add response-type=token to login/oauth/authorize for single page apps

cadecairos:
#356 

R?

It doesn't say that a client shouldn't be able to use it in any way - should we add that in? how should we do it? a new column?

### DIFF
--- a/docs/oauth.md
+++ b/docs/oauth.md
@@ -60,7 +60,7 @@ To start the web application flow you need to redirect your users browser to Web
 
 **Example**
 
-`GET https://id.webmaker.org/login/oauth/authorize?client_id=wm_id_BTQNPABtUqqApaDrcsDa&response_type=code&scope=user&state=Nvqfc67z`
+`GET https://id.webmaker.org/login/oauth/authorize?client_id=wm_id_BTQNPABtUqqApaDrcsDa&response_type=code&scopes=user&state=Nvqfc67z`
 
 ### 2. Webmaker redirects back to your application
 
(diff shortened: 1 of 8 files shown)


cadecairos:
@jbuck Added in allowed_responses to verify the client can indeed use the provided response_type

lib/queries.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
       "access_tokens WHERE access_token = $1;"
   },
   create: {
-    client: "INSERT INTO clients VALUES($1, $2, $3, $4);",
+    client: "INSERT INTO clients VALUES($1, $2, $3, $4, $5);",
```
jbuck: Could you name these columns explicitly? I worry about the case where the db column order isn't the same as what this code is expecting. If you wanna fix up the two lines below this too that'd be lovely :)

web/server.js
```diff
@@ -163,6 +163,19 @@ module.exports = function(options) {
       method: 'GET',
       path: '/login/oauth/authorize',
       config: {
+        validate: {
```
jbuck: mmmmmm tasty validations

web/server.js
```diff
@@ -163,6 +163,19 @@ module.exports = function(options) {
...
+          query: {
+            client_id: Joi.string().required(),
+            response_type: Joi.string().valid('code', 'token'),
+            scopes: Joi.string().required(),
+            state: Joi.string().when('response_type', {
```
jbuck: Let's make state a require parameter, to be consistent

web/server.js
```diff
@@ -163,6 +163,19 @@ module.exports = function(options) {
...
+              is: 'code',
+              then: Joi.required(),
+              otherwise: Joi.forbidden()
+            }),
+            action: Joi.string().optional().valid('signup')
```
jbuck: I think this should be `.default('signin')` right?

web/server.js
```diff
@@ -190,36 +203,49 @@ module.exports = function(options) {
...
-        redirectObj.query.client_id = request.query.client_id;
-        var redirectUri = url.format(redirectObj);
-
-        reply.redirect(redirectUri);
+        var expiresAt = new Date(Date.now() + 60 * 1000).toISOString();
+        var redirectObj = url.parse(client.redirect_uri, true);
+
+        scopes = scopes.split(' ');
+
+        oauthDb.generateAuthCode(
```
jbuck: Lets move this generateAuthCode function back into a pre method, and put the generateToken function in a pre method as well. We can have them conditionally run based on the response_type, and also return an error if it's not allowed. Then the handler can just be output formatting!

jbuck:
r-, needs some code fixings

docs/oauth.md
```diff
@@ -258,6 +258,51 @@ To access the API using your access token, set the `Authorization` header for ea
...
+  <tr>
+    <td>response_type</td>
+    <td>The response_type must be set to "token".</td>
+  </tr>
+  <tr>
+    <td>action</td>
+    <td>Optional. If provided, should be <code>signup</code> or <code>signin</code>. Send to improve the user experience, based on whether the user clicked on a Sign In or Sign Up button. <code>signin</code> is the default.</td>
+  </tr>
+  <tr>
+    <td>scopes</td>
```
jbuck: Can you add the `state` parameter to the docs here?

jbuck:
@cadecairos r+

PR #368, opened 2015-08-18
Login caching

### LINKED ISSUE #367: Add a cache layer between id and login.wmo
We hit the login server with an account lookup every time we verify a token, to get the account data.

we can probably safely add a cache segment there to get some quick and easy performance wins.

cadecairos:
Fixes #367 

Two commits here:

The first commit refactors the "account.js" file into a Hapi module that exposes its methods as Hapi server methods (this makes adding caching where we want it much easier) The methods don't change, only how they're called and required into the server.

The second commit adds in the caching to getUser, cached for 24 hours, stale in one, with a staleTimeout of 100ms (return cached value if refresh takes any longer). If there's an error or no user is returned, the result is not cached by passing "0" into the callback as the third argument.

### DIFF
--- a/lib/account.js
+++ b/lib/account.js
@@ -158,6 +158,7 @@ exports.register = function(server, options, next) {
   });
 
   server.method('account.getUser', function(userId, callback) {
+    var ttl = null;
     var getRequest = hyperquest({
       method: 'GET',
       headers: {
@@ -166,13 +167,26 @@ exports.register = function(server, options, next) {
       uri: loginAPI + '/user/id/' + userId
     });
 
-    getRequest.on('error', callback);
+    getRequest.on('error', function(err) {
+      callback(err, null, 0);
+    });
 
     getRequest.on('response', function(message) {
       parseMessage(message, function(err, json) {
-        callback(err, json);
+        if (json && !json.user) {
+          ttl = 0;
+        }
+
+        callback(err, json, ttl);
       });
     });
+  }, {
+    cache: {
+      segment: 'accounts.getUser',
+      expiresIn: 1000 * 60 * 60 * 24,
+      staleIn:  1000 * 60 * 60,
+      staleTimeout: 100
+    }
   });
 
   server.method('account.requestMigrateEmail', function(request, callback) {
--- a/lib/account.js
+++ b/lib/account.js
@@ -2,269 +2,281 @@ var Boom = require('boom');
 var hyperquest = require('hyperquest');
 var url = require('url');
 
-module.exports = function(config) {
-  // https://basic:auth@login.server.org
-  var loginAPI = config.loginAPI;
-
-  function getIPAddress(request) {
-    // account for load balancer!
-    if (request.headers['x-forwarded-for']) {
-      return request.headers['x-forwarded-for'];
-    }
-
-    return request.info.remoteAddress;
+function getIPAddress(request) {
+  // account for load balancer!
+  if (request.headers['x-forwarded-for']) {
+    return request.headers['x-forwarded-for'];
   }
 
-  function parseMessage(message, callback) {
-    var bodyParts = [];
-    var bytes = 0;
+  return request.info.remoteAddress;
+}
 
-    message.on('data', function(c) {
-      bodyParts.push(c);
-      bytes += c.length;
-    });
+function parseMessage(message, callback) {
+  var bodyParts = [];
+  var bytes = 0;
 
-    message.on('end', function() {
-      var body = Buffer.concat(bodyParts, bytes).toString('utf8');
-      var json;
+  message.on('data', function(c) {
+    bodyParts.push(c);
+    bytes += c.length;
+  });
 
-      if ( message.statusCode !== 200 ) {
-        try {
-          json = JSON.parse(body);
-        } catch (ex) {
-          return callback(Boom.create(message.statusCode, 'LoginAPI error', body));
-        }
-        return callback(Boom.create(message.statusCode, 'LoginAPI error', json));
-      }
+  message.on('end', function() {
+    var body = Buffer.concat(bodyParts, bytes).toString('utf8');
+    var json;
 
+    if ( message.statusCode !== 200 ) {
       try {
         json = JSON.parse(body);
       } catch (ex) {
-        return callback(Boom.badImplementation('Error parsing response from Login server', ex));
+        return callback(Boom.create(message.statusCode, 'LoginAPI error', body));
       }
-
-      callback(null, json);
-    });
-  }
-
-  return {
-    verifyPassword: function(request, callback) {
-      var loginRequest = hyperquest({
-        method: 'POST',
-        headers: {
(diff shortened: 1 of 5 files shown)


lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
       try {
         json = JSON.parse(body);
       } catch (ex) {
-        return callback(Boom.badImplementation('Error parsing response from Login server', ex));
+        return callback(Boom.create(message.statusCode, 'LoginAPI error', body));
       }
+      return callback(Boom.create(message.statusCode, 'LoginAPI error', json));
+    }
+
+    try {
```
humphd: Why do this in two spots, since you seem to need the JSON parsed no matter the `statusCode`.  Why not do it above 27?

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/verify-password'
+    loginRequest.on('error', callback);
+
+    loginRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: Don't wrap in a function here, just `parseMessage(message, callback);`

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-
-      var resetRequest = hyperquest({
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/request-reset-code'
+    resetRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: Same thing here, wrapping adds nothing in this case.

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-    resetPassword: function(request, callback) {
-      var resetRequest = hyperquest({
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/reset-password'
+    resetRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: And here

lib/account.js
cadecairos: Yeah, not sure why it ended up like this. We can file a bug to clean it up a bit.

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-    createUser: function(request, callback) {
-      var createRequest = hyperquest({
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/create'
+    createRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: And here.

lib/account.js
cadecairos: filed #396

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-
-      var migrateRequest = hyperquest({
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/request'
+    migrateRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: And here

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-    verifyToken: function(request, callback) {
-      var verifyRequest = hyperquest({
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/authenticateToken'
+    verifyRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: And here.

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-    setPassword: function(request, uid, password, callback) {
-      var passwordRequest = hyperquest({
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/enable-passwords'
+    passwordRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: And here

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-    checkUsername: function(request, callback) {
-      var usernameRequest = hyperquest({
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'x-ratelimit-ip': getIPAddress(request)
-        },
-        uri: loginAPI + '/api/v2/user/exists'
+    usernameRequest.on('response', function(message) {
+      parseMessage(message, function(err, json) {
```
humphd: And here

lib/account.js
cadecairos: sure can: filed #397

web/server.js
```diff
@@ -18,15 +18,26 @@ var passTest = new PassTest({
...
         security: true
       }
     }
-  });
+  };
+
+  if ( options.redisUrl ) {
+    var redisUrl = require('redis-url').parse(options.redisUrl);
+    serverConfig.cache = {
+      engine: require('catbox-redis'),
```
jbuck: You need to add this dependency in package.json, but I think we should consider offering catbox-memory as a default solution too.

lib/account.js
```diff
@@ -2,269 +2,295 @@ var Boom = require('boom');
...
-          'Content-Type': 'application/json'
-        },
-        uri: loginAPI + '/user/id/' + userId
+        callback(err, json, ttl);
       });
+    });
+  }, {
+    cache: {
+      segment: 'accounts.getUser',
+      expiresIn: 1000 * 60 * 60 * 24,
```
jbuck: Lets set this to something shorter like 60 seconds (so 60 * 1000), and staleIn to 30 seconds. That should be long enough for a burst series of requests.

jbuck:
@cadecairos this looks pretty good! I made some notes around cache timing and making the developer experience better

web/server.js
cadecairos: if no redisUrl is specified, hapi automagically uses catbox as the default server cache engine. ITS MAGICALLY DELICIOUS

jbuck:
@cadecairos r+

Issue #361, opened 2015-08-12
No indication of successful password reset


vazquez:
We need to show a visual indicator upon successful reset of the user's password.

**Message if user comes from Webmaker (or Teach, or any other app):** "Success! Your password has been reset. If you are not automatically sent back to Webmaker, [click here](#)."

**Message if user comes from id.webmaker.org:** "Success! Your password has been reset. Sign in with your new password below."

#### Designs

![1](https://cloud.githubusercontent.com/assets/1119821/9232893/30eb165c-40fe-11e5-8dd2-aab74aa954a3.png)  

![2](https://cloud.githubusercontent.com/assets/1119821/9232892/30e50050-40fe-11e5-93ea-6754f0c5ed5f.png)

vazquez:
Resolving this issue helps in fixing mozilla/webmaker-core#441

xmatthewx:
@cadecairos - ideally, we show the message then redirect users after about 3 seconds.

cadecairos:
We're gonna need someone to do the HTML and CSS for these - adding the logic to redirect shouldn't be hard
Points of interest:
add message before redirection to android app here: https://github.com/mozilla/id.webmaker.org/blob/e89b16d4de94677b027c8521f250d5999f3b5df9/templates/pages/reset-password.jsx#L84-L101
add message for desktop based resets here: https://github.com/mozilla/id.webmaker.org/blob/e89b16d4de94677b027c8521f250d5999f3b5df9/templates/pages/login.jsx#L60

cadecairos:
(I've got no idea how to make these look like the mockups)

xmatthewx:
I believe @vazquez followed the style of existing notices in there. Regardless, these don't need to be fancy, just functional and not ugly.

vazquez:
CSS is exactly the same, just the `fa` icons are changed, along with the
text.

Ricardo Vazquez
UI Designer, Webmaker
Mozilla Foundation
@iamrvazquez <http://twitter.com/iamrvazquez>
IRC: ricardo

PR #369, opened 2015-08-20
display success message on a successful password reset

cadecairos:
WIP: The android reset message isn't displaying properly still.

#361 

screenshot from 2015-08-20 17-09-02

PR #370, opened 2015-08-21
Expired Reset Message

### LINKED ISSUE #360: Expired token message
When a user visits the reset password page after a token has been used / expired, we do not provide any message that this token is expired.

Message: "Seems that the link in your email expired. [Reset your password again](#)."

This is how it could look like, following the project's already implemented alerts (icon being used is `fa-warning`)

screenshot 2015-08-12 14 04 52

cadecairos:
Fixes #360 

Land after #361

### DIFF
--- a/templates/components/password-reset-expired.jsx
+++ b/templates/components/password-reset-expired.jsx
@@ -9,6 +9,9 @@ var PasswordResetExpired = React.createClass({
         iconClass="messageIcon fa fa-exclamation-triangle"
         header="Seems that the link in your email expired."
         headerClass="messageHeader">
+          <p>
+            Request a new reset password link below
+          </p>
       </IconText>
     );
   }
--- a/templates/components/password-reset-expired.jsx
+++ b/templates/components/password-reset-expired.jsx
@@ -1,0 +1,17 @@
+var React = require('react');
+var IconText = require('./icontext.jsx');
+
+var PasswordResetExpired = React.createClass({
+  render: function() {
+    return (
+      <IconText
+        className="messageBox"
+        iconClass="messageIcon fa fa-exclamation-triangle"
+        header="Seems that the link in your email expired."
+        headerClass="messageHeader">
+      </IconText>
+    );
+  }
+});
+
+module.exports = PasswordResetExpired;
--- a/templates/components/password-reset-expired.jsx
+++ b/templates/components/password-reset-expired.jsx
@@ -7,10 +7,10 @@ var PasswordResetExpired = React.createClass({
       <IconText
         className="messageBox"
         iconClass="messageIcon fa fa-exclamation-triangle"
-        header="Seems that the link in your email expired."
+        header="The link in your email expired."
         headerClass="messageHeader">
           <p>
-            Request a new reset password link below
+            Request a new link below to set your password
           </p>
       </IconText>
     );
--- a/templates/lib/utils.js
+++ b/templates/lib/utils.js
@@ -1,0 +1,3 @@
+exports.getClassNameString = function() {
+  return Array.prototype.slice.call(arguments).join(" ");
+};
--- a/templates/lib/utils.js
+++ b/templates/lib/utils.js
@@ -1,3 +1,0 @@
-exports.getClassNameString = function() {
-  return Array.prototype.slice.call(arguments).join(" ");
-};
--- a/templates/pages/login.jsx
+++ b/templates/pages/login.jsx
@@ -1,5 +1,6 @@
 var React = require('react');
 var Router = require('react-router');
+var classnames = require('classnames');
 var Link = Router.Link;
 
 var Form = require('../components/form/form.jsx');
@@ -56,11 +57,9 @@ var Login = React.createClass({
     var buttonText = "Log In";
     this.queryObj = Url.parse(window.location.href, true).query;
 
-    var wrapperClass = "centerDiv";
-
-    if (this.queryObj.passwordReset) {
-      wrapperClass = utils.getClassNameString(wrapperClass, "largeWrapper");
-    }
+    var wrapperClass = classnames("centerDiv", {
+      largeWrapper: !!this.queryObj.passwordReset
+    });
 
     return (
       <div>
--- a/templates/pages/login.jsx
+++ b/templates/pages/login.jsx
@@ -6,7 +6,6 @@ var Link = Router.Link;
 var Form = require('../components/form/form.jsx');
 var Header = require('../components/header/header.jsx');
 var PasswordResetSuccess = require('../components/password-reset-success.jsx');
-var utils = require('../lib/utils.js');
 var WebmakerActions = require('../lib/webmaker-actions.jsx');
 var Url = require('url');
 var ga = require('react-ga');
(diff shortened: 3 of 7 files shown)


templates/pages/reset-password.jsx
```diff
@@ -32,28 +35,52 @@ var ResetPassword = React.createClass({
     linkQuery.state = this.state.queryObj.state;
     linkQuery.scopes = this.state.queryObj.scopes;
     linkQuery.response_type = this.state.queryObj.response_type;
+
+    var wrapperClass = "centerDiv";
```
Pomax: possible nit: looks like this does the same class assignment as PasswordResetSuccess, worth refactoring to a utility function that both can call?

templates/components/password-reset-expired.jsx
```diff
@@ -0,0 +1,20 @@
...
+var PasswordResetExpired = React.createClass({
+  render: function() {
+    return (
+      <IconText
+        className="messageBox"
+        iconClass="messageIcon fa fa-exclamation-triangle"
+        header="Seems that the link in your email expired."
+        headerClass="messageHeader">
+          <p>
+            Request a new reset password link below
```
Pomax: perhaps "Request a new reset link for setting a new password below", to mitigate possible confusion around saying "request a link" here, but "set a password" on the actual on-page button? (see https://github.com/mozilla/id.webmaker.org/issues/360#issuecomment-134677659)

Pomax:
pretty much R+, although the phrasing might need a tweak before landing this (and of course a rebase after #369 lands too)

templates/components/password-reset-expired.jsx
cadecairos: WFM, @xmatthewx does that sound good to you?

templates/pages/reset-password.jsx
cadecairos: Not sure if it's worth it

```js
exports.getClassNameString = function() {
  return Array.prototype.slice.call(arguments).join(" ");
};

```

```js
if (this.state.expired) {
  wrapperClass = utils.getClassNameString("resetPasswordPage", "centerDiv", "largeWrapper");
} else {
  wrapperClass = utils.getClassNameString("resetPasswordPage", "centerDiv");
}
```

cadecairos: I suppose you could also do

```js
var wrapperClass = "resetPasswordPage centerDiv";

if(this.state.expired) {
    wrapperClass = utils.getClassNameString(wrapperClass, "largeWrapper");
}
```

templates/components/password-reset-expired.jsx
xmatthewx: Good fix @Pomax. Following your lead with a few less words:

"Request a new link below to set your password"

templates/components/password-reset-expired.jsx
```diff
@@ -0,0 +1,20 @@
+var React = require('react');
+var IconText = require('./icontext.jsx');
+
+var PasswordResetExpired = React.createClass({
+  render: function() {
+    return (
+      <IconText
+        className="messageBox"
+        iconClass="messageIcon fa fa-exclamation-triangle"
+        header="Seems that the link in your email expired."
```
xmatthewx: Generally I like casual language, but this is likely a frustrating and confusing moment for users. Let's be more clear and concise:

"The link in your email expired."

templates/lib/utils.js
```diff
@@ -0,0 +1,3 @@
+exports.getClassNameString = function() {
```
Pomax: this seems really poorly named. If the only thing this does is join any iterable with spaces, then let's not call it `getClassNameString`, or a file called `utils.js`, neither of which reflects what it's actually doing =)

templates/pages/login.jsx
```diff
@@ -58,7 +59,7 @@ var Login = React.createClass({
     var wrapperClass = "centerDiv";
 
     if (this.queryObj.passwordReset) {
-      wrapperClass += " largeWrapper"
+      wrapperClass = utils.getClassNameString(wrapperClass, "largeWrapper");
```
Pomax: at this point, I'm going to say "just use `classnames` instead", because writing shim code ourselves is a little silly.
```
var classnames = require('classnames');

var className = classnames({
  centerDiv: true,
  largeWrapper: !!this.queryObj.passwordReset
});
```

templates/pages/reset-password.jsx
```diff
@@ -36,6 +40,10 @@ var ResetPassword = React.createClass({
     linkQuery.scopes = this.state.queryObj.scopes;
     linkQuery.response_type = this.state.queryObj.response_type;
 
+    if (this.state.expired) {
```
Pomax: definitely use `classnames`

templates/pages/reset-password.jsx
```diff
@@ -121,10 +133,20 @@ var ResetPassword = React.createClass({
...
         return;
       }
-      // handle errors!
+
+      this.setState({
+        expired: true,
+        submitForm: false,
+        email: false
+      });
     }.bind(this)).catch(function(ex) {
```
Pomax: is there support for `.finally(function(...) { ... })`, here, so we can house that duplicated `this.setState` in a single block?

templates/pages/login.jsx
```diff
@@ -1,10 +1,12 @@
 var React = require('react');
 var Router = require('react-router');
+var classnames = require('classnames');
 var Link = Router.Link;
 
 var Form = require('../components/form/form.jsx');
 var Header = require('../components/header/header.jsx');
 var PasswordResetSuccess = require('../components/password-reset-success.jsx');
+var utils = require('../lib/utils.js');
```
Pomax: we no longer use this now, right?

templates/pages/reset-password.jsx
```diff
@@ -1,10 +1,13 @@
 var React = require('react');
+var classnames = require('classnames');
 
 var Header = require('../components/header/header.jsx');
 var IconText = require('../components/icontext.jsx');
 var ResetView = require('../components/reset-password-view.jsx');
 var RequestView = require('../components/request-reset-view.jsx');
 var PasswordResetSuccess = require('../components/password-reset-success.jsx');
+var PasswordResetExpired = require('../components/password-reset-expired.jsx');
+var utils = require('../lib/utils.js');
```
Pomax: no longer used?

templates/pages/reset-password.jsx
```diff
@@ -121,10 +139,12 @@ var ResetPassword = React.createClass({
         window.location = Url.format(redirectObj);
         return;
       }
-      // handle errors!
+
+      this.expire();
     }.bind(this)).catch(function(ex) {
```
Pomax: Would this work?
```
      ... // no this.expire() as last call
     }.bind(this)).catch(function(ex) {
       ... // no this.expire() as last call here, either
    }).then(this.expire);
```

templates/pages/reset-password.jsx
```diff
@@ -121,10 +139,12 @@ var ResetPassword = React.createClass({
...
       ga.event({category: 'Reset Password', action: 'Error parsing response from the server'});
       console.error('Error parsing response', ex);
+      this.expire();
     });
```
Pomax: if we do keep this.expire() here, we probably need another `.bind(this)` here

xmatthewx:
reassign to @cadecairos ?

PR #371, opened 2015-08-24
Add a simple android UA check before redirecting to webmaker://

### LINKED ISSUE #359: password reset via email for webmaker app directs to webmaker://login?mode=sign-in
That protocol only works on android devices, so on desktop (which users may very well be on when they handle the emailed reset option) it'll look like the password reset horribly failed. If there is a way to even sort-of guess whether we're on "not android" so we can redirect to a "your password was reset, go do your thing now" page instead of redirecting into the webmaker app, that sounds like it'll be worth doing.

cadecairos:
Fixes #359

### DIFF
--- a/templates/pages/reset-password.jsx
+++ b/templates/pages/reset-password.jsx
@@ -87,7 +87,11 @@ var ResetPassword = React.createClass({
         queryObj = Url.parse(window.location.href, true).query;
         ga.event({category: 'Reset Password', action: 'Successfully request new password'});
         if ( queryObj.android === 'true' ) {
-          if (navigator.userAgent.toLowerCase().indexOf("android") > -1) {
+          if (
+            navigator &&
+            navigator.userAgent &&
+            navigator.userAgent.toLowerCase().indexOf("android") > -1
+          ) {
             window.location = "webmaker://login?mode=sign-in";
             return;
           }
--- a/templates/pages/reset-password.jsx
+++ b/templates/pages/reset-password.jsx
@@ -87,8 +87,10 @@ var ResetPassword = React.createClass({
         queryObj = Url.parse(window.location.href, true).query;
         ga.event({category: 'Reset Password', action: 'Successfully request new password'});
         if ( queryObj.android === 'true' ) {
-          window.location = "webmaker://login?mode=sign-in";
-          return;
+          if (navigator.userAgent.toLowerCase().indexOf("android") > -1) {
+            window.location = "webmaker://login?mode=sign-in";
+            return;
+          }
         }
         redirectObj = Url.parse('/login', true);
         redirectObj.query.client_id = queryObj.client_id;


templates/pages/reset-password.jsx
```diff
@@ -87,8 +87,10 @@ var ResetPassword = React.createClass({
         queryObj = Url.parse(window.location.href, true).query;
         ga.event({category: 'Reset Password', action: 'Successfully request new password'});
         if ( queryObj.android === 'true' ) {
-          window.location = "webmaker://login?mode=sign-in";
-          return;
+          if (navigator.userAgent.toLowerCase().indexOf("android") > -1) {
```
Pomax: just to make sure this doesn't break unexpectedly, maybe worth expanding the conditional a little?
```
if (navigator && navigator.userAgent && navigator.userAgent.toLow[...]) {
...
}
```

cadecairos: Which browsers don't have window.navigator.userAgent?

afaict, it's supported pretty much everywhere: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/userAgent

Pomax: why not protect against breaking for people who use extensions/addons that kill off the user agent property?

cadecairos:
@Pomax I think we need to change how this handles resets *originating* from Webmaker Android, but *completed* on desktop (`android=true` flag in the reset link)

Those reset links won't contain any OAuth 2 params, and won't actually let the user sign in on the page we send them to. Think about opening that little google OAuth popup box directly and trying to sign in, same deal here.

We should not do an automatic redirect, But display a success message. I'm not sure about the copy there.

@xmatthewx what do you think?

xmatthewx:
To clarify, does this only impact A) people who reset in a browser on a device other than their Android? We're not talking about B) people who reset in their Android browser, correct?

If so, good catch. Are you suggesting that we don't do redirect for anyone, group A or group B?
 
Most users won't have this experience, so we shouldn't degrade group B's experience. If we can improve group A independently, that's great. I'll craft a success message.

cadecairos:
This is an issue for users who start a reset flow from the android app, and complete the reset flow in a desktop browser, rather than a mobile browser.

xmatthewx:
Ok. If we can address that group independently, how about:

> Success! Your password has been set.
Open the app to login.

I would say this group includes desktop browser & iOS devices, or anything non-Android.

cadecairos:
> I would say this group includes desktop browser & iOS devices, or anything non-Android.

I'll trigger it if "android" is not in the browser user agent, but `android=true` is part of the link used to reset the password

Pomax:
To be fair: I don't have an email client on my mobile device, so if I reset my password on device, I will still use my desktop to actually pick up my mail and follow the link it sent me. Can we just give two links in the email? One for "if you're on desktop, click here" and one for "if you're on android, click here"?

cadecairos:
But where would the for-desktop link go? beta.webmaker.org?

xmatthewx:
What's wrong with this solution?

> I'll trigger it if "android" is not in the browser user agent, but android=true is part of the link used to reset the password

Two links will confuse the majority of people who have a single device for all the things. And, most people with two devices will still click the fat reset button instead of reading an alternative text link below.

Pomax:
that solution seemed reasonable to me, but "I'll trigger it if "android" is not in the browser user agent, but android=true is part of the link used to reset the password" made it sound like we were generating the link to visit with the "which device is going to be used" baked in, which is probably a bad idea.

Pomax:
R+ on the code as it is now, btw (aside from the rebase of course)

PR #377, opened 2015-09-01
Only add the required querystring parameters

jbuck:
The OAuth authorization route only requires the client_id,
response_type, scopes, and state parameters. The migration page would
send all of the parameters on its querystring so the authorization page
throws a validation error.


jbuck:
@ashleygwilliams r?

jbuck:
Fix https://github.com/mozilla/thimble.webmaker.org/issues/995

ashleygwilliams:
looks good to me! r+

PR #383, opened 2015-09-09
there is no phantomjs 1.9.7-15, it goes from 1.9.7-14 to 1.9.8

Pomax:
The phantom module as required right now fails because there is no `1.9.7-15`, so nothing gets installed and mocha-phantom dies trying to use a non-existent version of Phantom.


cadecairos:
R+

PR #384, opened 2015-09-11
added an OPTIONS handler to deal with the CORS check for /user

Pomax:
without the OPTIONS entry for the /user route, the browser will fail its preflight CORS check on the /user route, and will never allow the GET operation to go through


cadecairos:
R+

Issue #318, opened 2015-06-08
Evaluate password strength requirements 


xmatthewx:
It took me 15 minutes to set a password with enough entropy to satisfy Webmaker. It seems like 14 characters with a mix of uppercase, lowercase, numbers and a weird symbol or two should be enough for the accounts we are trying to protect. I knew what was happening but still wondered if the validation was broken. 

Can we dial this back a little bit? I worry that we will lose potential users on Android. 

cc @secretrobotron @simonwex

thisandagain:
I'm seeing this as a regular frustration with users as well. I suggest we add promote this to launch blocker. Propose replacing rules with:

- Minimum password length: 6 characters
- Maximum password length: 256 characters

Remove all other constraints. If we want to protect users we should do it through throttling API requests and monitoring not by throwing up UX hurdles to the point where users end up writing passwords down on sticky notes. For those interested, there is a password management user research brown bag on Thursday that will address some of this. /cc @simonwex

simonwex:
Totally agree with @thisandagain. The only thing that a more complicated password system does is make people write them down on post-its for others to see.

ashleygwilliams:
:+1: i also agree.

the best thing i've found is making the suggestion that pass*phrases* are very safe. sentences are easier to remember than garbage characters and numbers, capital or not- they are also *way* safer. people don't even consider it an option unless you mention it.

https://xkcd.com/936/
![screenshot 2015-06-09 at 8 05 49 pm](https://cloud.githubusercontent.com/assets/1163554/8072367/f49b1d20-0ee2-11e5-8934-807bc26fb053.png)

cadecairos:
> Minimum password length: 6 characters

I disagree, minimum should be 8, and even then that is considered weak (https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Password_Length)

If we're to remove other limitations we need at least two patches:

1. login.webmaker.org, to remove serverside strength checks
2. id.webmaker.org to remove server and client side strength

We can go even further and update webmaker-login-ux and any apps using it for login.

We should actively be encouraging passphrases if we're not enforcing other rules. I would also prefer if we continued disallowing the 10k most commonly used passwords

xmatthewx:
If someone uses a common password, let's tell them and teach them (if we aren't already). Error : "That is a very common password and easy to guess. Try something else."

cadecairos:
[...]

PR #388, opened 2015-09-15
Added a basic password strength meter

matthewsot:
See issue #318 and https://github.com/mozilla/api.webmaker.org/issues/178, using zxcvbn to add a password meter instead of specific character requirements. Anything with a score < 2/4 will be blocked from signing up.

I haven't been able to test actually creating a new account with this, and it can be a little bit buggy (especially if you click the Password field then click away then enter a password, it doesn't recognize that a password has been entered until you click away [though I think it's always been like that?])

Also note that I've left the check for passwords to be at least 8 characters, although I would suggest changing that to 6 characters as long as zxcvbn claims the password is secure enough.

0/4 Password
1/4 Password
2/4 Password
3/4 Password
4/4 Password

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
     "mofo-style": "latest",
     "ncp": "^2.0.0",
     "npm-run-all": "^1.1.3",
-    "phantomjs": "1.9.7-15"
+    "phantomjs": "1.9.7-15",
+    "zxcvbn": "^3.4.0"
   },
   "engines": {
     "npm": "^2.5.1",
--- a/templates/components/form/form.jsx
+++ b/templates/components/form/form.jsx
@@ -121,6 +121,11 @@ var Form = React.createClass({
     });
 
   },
+  handleKeyUp: function(event) {
+   if ( this.props.onInputKeyUp ) {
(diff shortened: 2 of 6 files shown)


jbuck:
@ashleygwilliams @cadecairos r?

jbuck:
Also, we'll need to do the same patch on the server-side

matthewsot:
Also realized that this might not have updated the reset-password page as well, need to test that

cadecairos:
@matthewsot This is really great stuff, and very much appreciated!

I do have a couple concerns though. 

Firstly, It seems a bit too strict. While I'm all for super strong passwords, zxcvbn is rejecting `CheesyPoof1` which, while not the most sophisticated password, I would consider it "good enough". Basically, while the addition of a strngth bar is great, it seems that at the same time we've made it harder for average users to select memorable passwords for themselves. 

Secondly, the zxcvbn package is [320kb, bundled and gzipped](https://github.com/dropbox/zxcvbn#script-load-latency) - a ridiculously large package to load, especially on a 3g mobile network... This is one of the chief reasons I didn't go with zxcvbn when choosing a password strength checker. We need to pull in more folks to weigh in on the pros and cons of loading such a huge package.

PR #389, opened 2015-09-15
Only add required parameters for OAuth authorization, part 2

jbuck:
We got an email from Sean saying he ran into this again... so I decided to fix this everywhere.


jbuck:
@ashleygwilliams @cadecairos r?

PR #392, opened 2015-09-16
return when the call to login errors in GET /user

### LINKED ISSUE #391: Don't reply() twice in GET /user
https://github.com/mozilla/id.webmaker.org/blob/develop/web/server.js#L582-L586 should have a `return` in the error case so you don't double call `reply()`.

cc @cadecairos

cadecairos:
Fixes #391 

@jbuck R?


jbuck:
@cadecairos r+

PR #403, opened 2015-10-27
Add user DB to ID server

cadecairos:
This is quite a large patch, but necessarily so. It brings the user, password, and password reset code storage over from login.wmo.

In order to maintain user ids from the older system, the user tables' sequence table starts at 1,000,000. We can then import users from a login db and then we can add foreign key relations, and new users ids won't have collisions with old ones.

migration flow is:

1. run `psql -d <oauthDB> -f migrations/1_add_user_tables.sql`
2. run `node scripts/migrate_legacy_webmaker_users.js`
3. run `psql -d <oauthDB> -f migrations/2_add_user_fk.sql`

Here's an overview of the changes:

* Rewrote all the queries in Knex
* Refactored the OAuthDB module to use the knex code, made the code simpler and easier to read.
* I moved all the crypto related functions into a separate module

Things to note:
* I had to create a custom jshintrc so that jshint stopped bugging me about ES6 features. I should probably open up an upstream issue to add support for stable ES6 features (arrow notation, block scoped variables, object property shorthand, etc).
*  I still have to write some tests, and do an audit of the stuff I've written, to make sure this patch is at pairity with when it connected to the wmlogin server.


humphd:
You can't hear it, but I'm clapping.  This is a huge perf win for consumers of the ID server.  Nice work. cc @gideonthomas

lib/CryptoUtils.js
```diff
@@ -0,0 +1,72 @@
...
+        resolve(salt);
+      });
+    });
+  }
+
+  hash(password, salt) {
+    return new Promise((resolve, reject) => {
+      bcrypt.hash(password, salt, (err, hash) => {
+        if (err) {
+          reject(err);
```
humphd: Missing `return`

cadecairos:
bcrypt 8.5.0 doesn't seem to be able to compile on the travis build machines...

Since speed isn't really necessary when dealing with bcrypt, I'll look into pure JS implementations that have a solid community of contributors

lib/CryptoUtils.js
cadecairos: aha! nice catch :)

migrations/1_add_user_tables.sql
```diff
@@ -0,0 +1,25 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS "users"
+(
+  id bigserial NOT NULL,
+  email varchar NOT NULL UNIQUE,
+  username varchar NOT NULL UNIQUE,
```
humphd: You should consider adding indices on the `email` and `username` columns, since you query on those above.

scripts/create-tables.sql
```diff
@@ -1,3 +1,25 @@
+CREATE TABLE IF NOT EXISTS "users"
+(
+  id bigserial NOT NULL,
+  email varchar NOT NULL UNIQUE,
+  username varchar NOT NULL UNIQUE,
```
humphd: I guess it's probably in here vs the migration file above, but same applies: index columns you search on.

migrations/1_add_user_tables.sql
cadecairos: yes, lets do this

cadecairos:
@jbuck R? plz

cadecairos:
Also, I'm open to separating out the creation and population of the user db into a separate patch, after which we can ship the rest.

PR #404, opened 2015-10-30
auth:false for OPTIONS

Pomax:
Hapi 11 changed how OPTIONS preflight calls are handled, and the explicit configuration might be throwing things off.


cadecairos:
:ship: it

PR #415, opened 2016-02-23
Use a pre-eslint version of mofo-style

cadecairos:
ESLinting provided by the latest mofo-style breaks us. setting a hard dependency on v1.1.0


alicoding:
r+

alicoding:
travis is failing.

cadecairos:
Yeah I can't get it to fail locally though. looks like a problem with phantomjs and travis

alicoding:
merge it if you think this is travis only

PR #428, opened 2017-02-10
Add pre-install script to copy env before install

ryanwarsaw:
This verifies that the .env file has been copied before trying to run the post-install script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@
 1. Fork and clone this repository
 2. Navigate to the directory of the repository, e.g. `cd id.webmaker.org`
 3. `npm install` to install dependencies
-4. Set configuration variables, by default `cp sample.env .env` or `copy sample.env .env` on Windows
 5. `npm start`
 6. Navigate your browser to [`http://localhost:1234`](http://localhost:1234)
 
--- a/scripts/ensure-env.js
+++ b/scripts/ensure-env.js
@@ -1,0 +1,5 @@
+var fs = require("fs");
+var env = require("path").join(__dirname, "..", ".env");
+
+if (fs.existsSync(env) || process.env.NODE_ENV == "production") process.exit(0);
+fs.createReadStream("sample.env").pipe(fs.createWriteStream(env));
--- a/scripts/ensure-env.js
+++ b/scripts/ensure-env.js
@@ -1,5 +1,5 @@
 var fs = require("fs");
 var env = require("path").join(__dirname, "..", ".env");
 
-if (fs.existsSync(env) || !(typeof process.env.NODE_ENV === 'undefined')) process.exit(0);
+if (fs.existsSync(env) || typeof process.env.NODE_ENV !== 'undefined') process.exit(0);
 fs.createReadStream("sample.env").pipe(fs.createWriteStream(env));
--- a/scripts/ensure-env.js
+++ b/scripts/ensure-env.js
@@ -1,5 +1,5 @@
 var fs = require("fs");
 var env = require("path").join(__dirname, "..", ".env");
 
-if (fs.existsSync(env) || process.env.NODE_ENV == "production") process.exit(0);
+if (fs.existsSync(env) || !(typeof process.env.NODE_ENV === 'undefined')) process.exit(0);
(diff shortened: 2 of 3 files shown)


scripts/ensure-env.js
```diff
@@ -0,0 +1,7 @@
+var fs = require("fs");
+var env = require("path").join(__dirname, "../", ".env");
```
Pomax: `join(__dirname, "..", ".env");` (path takes care of putting the correct delimiter in between the various path segments)

scripts/ensure-env.js
```diff
@@ -0,0 +1,7 @@
...
+
+fs.exists(env, function (exists) {
```
Pomax: We don't need async here, so keeping things synchronous actually yields a more predictable code path:
```
var exists = fs.existsSync(env);
if (exists) ...
...
```

README.md
```diff
@@ -17,7 +17,6 @@
 1. Fork and clone this repository
 2. Navigate to the directory of the repository, e.g. `cd id.webmaker.org`
 3. `npm install` to install dependencies
-4. Set configuration variables, by default `cp sample.env .env` or `copy sample.env .env` on Windows
```
Pomax: ❤️ :

ryanwarsaw:
cc: @gideonthomas @cadecairos R+?

Pomax:
R+ on my side

gideonthomas:
I'm pretty sure the `pre-install` script is run on heroku as well. Won't this then use what's in sample.env over what's in the actual env?

Pomax:
Easy enough to fix: this is a node script, so we can simply add an `if (process.env.NODE === "PRODUCTION") process.exit(0);`

scripts/ensure-env.js
```diff
@@ -0,0 +1,5 @@
+var fs = require("fs");
```
Pomax: based on https://github.com/mozilla/id.webmaker.org/pull/428#issuecomment-279095560 we'll need a check here to make sure we're on a local dev's machine, not on heroku...

ryanwarsaw:
@Pomax @gideonthomas Thanks for catching that, i went ahead and made the review changes requested, and learned a lot about Node's process API and environment-related stuff 👀  R+?

I also wanted to pick your brains on a question related to styling and JS, I'm used to working with languages like Java where there are several prominent standards (Google & Oracle's guides). However, with JS it seems to be much more diverse and spread out across the board. Is there a prominent standard with JS, or is it one of those things where as long as it passes lint it's okay?

🤔 💭

ryanwarsaw:
I just realized I have to check for ```staging``` as well, I'm going to do it in the morning since it's 1 AM and I can't think straight. 😴

Pomax:
@ryanwarsaw you can just check for a `NODE_ENV` var at all. If there is one, don't do anything, because whoever has that set clearly has special intentions.

ryanwarsaw:
@Pomax R+?

scripts/ensure-env.js
```diff
@@ -1,5 +1,5 @@
 var fs = require("fs");
 var env = require("path").join(__dirname, "..", ".env");
 
-if (fs.existsSync(env) || process.env.NODE_ENV == "production") process.exit(0);
+if (fs.existsSync(env) || !(typeof process.env.NODE_ENV === 'undefined')) process.exit(0);
```
Pomax: small nit: ` !(typeof process.env.NODE_ENV === 'undefined')` => `typeof process.env.NODE_ENV !== 'undefined'`, just to remove some parentheses.

scripts/ensure-env.js
```diff
@@ -0,0 +1,5 @@
+var fs = require("fs");
+var env = require("path").join(__dirname, "..", ".env");
+
+if (fs.existsSync(env) || !(typeof process.env.NODE_ENV === 'undefined')) process.exit(0);
```
Pomax: small nit: `!(typeof process.env.NODE_ENV === 'undefined')` => `typeof process.env.NODE_ENV !== 'undefined'`

ryanwarsaw:
@Pomax R+? Thanks for the reviews.

Pomax:
yup R+

PR #433, opened 2017-05-03
Updated images

flukeout:
Follow up to https://github.com/mozilla/id.webmaker.org/pull/432

* Mozilla logo in header no longer appears in all black in Firefox
* Split the checkmark image into two separate images because...
  * It needs to be green for the email opt-in checkbox
  * It needs to be white near the Terms of Service links

Should look like this...

image

cc @cadecairos


flukeout:
To test, go to the Sign Up page and observe the following:

image

flukeout:
Thanks @cadecairos

PR #435, opened 2017-05-11
Disables submit button on sign up / sign in forms.

flukeout:
Disables the sign up and log in buttons after they are clicked, re-enables when appropriate.
This change is to prevent people from potentially spamming the server with requests while waiting for a response.

Fixes https://github.com/mozilla/thimble.mozilla.org/issues/2068

cc @Pomax


Pomax:
This looks good to me

PR #447, opened 2018-01-12
Update to node 8

gideonthomas:
- Updated Hapi + related dependencies
- Extracted react-validation-mixin logic that we need and customize the mixin for use in the form on the front-end.


.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
```
gideonthomas: we used jshint and jscs before which have been deprecated and fail miserably on linting es8 so I had to switch to eslint. This eslint config is for the server and files that aren't in the client.

lib/account.js
```diff
@@ -28,9 +28,15 @@ function parseMessage(message, callback) {
       try {
         json = JSON.parse(body);
       } catch (ex) {
-        return callback(Boom.create(message.statusCode, 'LoginAPI error', body));
+        return callback(new Boom('LoginAPI error', {
```
gideonthomas: Boom API change here, `Boom.create` does not exist anymore.

I also kept these generic functions as callback-driven vs. switching to promises so as to minimize the number of unnecessary changes to software that we are going to deprecate.

lib/account.js
```diff
@@ -43,141 +49,171 @@ function parseMessage(message, callback) {
   });
 }
 
-exports.register = function(server, options, next) {
+const register = function(server, options) {
```
gideonthomas: change to the Hapi register API. See at the bottom of the file for what it has turned into.

lib/account.js
```diff
@@ -43,141 +49,171 @@ function parseMessage(message, callback) {
...
-    var loginRequest = hyperquest({
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json',
-        'x-ratelimit-ip': getIPAddress(request)
-      },
-      uri: loginAPI + '/api/v2/user/verify-password'
-    });
+  server.method('account.verifyPassword', function(request) {
+    return new Promise(function(resolve, reject) {
```
gideonthomas: all server methods now just return promises. No logic changes here otherwise.

lib/account.js
```diff
@@ -43,141 +49,171 @@ function parseMessage(message, callback) {
...
+        uri: loginAPI + '/user/id/' + userId
+      });
 
-    getRequest.on('response', function(message) {
-      parseMessage(message, function(err, json) {
-        if (json && !json.user) {
-          ttl = 0;
-        }
+      getRequest.on('error', function(err) {
+        flags.ttl = 0;
```
gideonthomas: Since server methods no longer support callbacks, you can't pass the ttl for the server method via the callback. Instead, according to Hapi docs, each server method receives a `flags` property that I believe you can set the ttl (and some other stuff) on. I'm pretty sure that's how it works, but no idea how to test it.

lib/oauth-db.js
```diff
@@ -32,118 +32,132 @@ function executeQuery(query, params, callback) {
   });
 }
 
-OAuthDB.prototype.getClient = function(clientId, callback) {
-  executeQuery(queries.get.client, [clientId], function(err, client) {
-    if ( err ) {
-      return callback(err);
-    }
+OAuthDB.prototype.getClient = function(clientId) {
```
gideonthomas: no logic changes in this file...only switched it to promises

ryanwarsaw:
You know you're next level when you not only write code, but review your own code too 😱

package.json
```diff
@@ -46,60 +44,60 @@
...
+    "eslint-plugin-react": "^7.5.1",
     "js-beautify": "1.5.5",
-    "jscs": "1.11.3",
-    "jshint": "2.6.3",
     "json-loader": "0.5.1",
     "lab": "13.1.0",
     "less": "2.4.0",
     "mocha": "2.2.1",
     "mocha-phantomjs": "3.6.0",
-    "mofo-style": "1.1.0",
```
gideonthomas: mofo-style eslint rules are nuts so I just use eslint:recommended

templates/.eslintrc.json
```diff
@@ -0,0 +1,22 @@
+{
```
gideonthomas: this file is specifically for front-end code so that eslint can lint react.

templates/components/validation-mixin.jsx
```diff
@@ -0,0 +1,123 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+---
+*/
+
+const factory = require('react-validation-mixin/lib/validationFactory').default;
```
gideonthomas: we only use the logic from the mixin and write our own component

gideonthomas: most of the code in this file is identical to the component in the actual mixin.

templates/components/validation-mixin.jsx
```diff
@@ -0,0 +1,123 @@
...
+   *
+   * @param {String|Function} key to validate, or error-first containing the validation errors if any.
+   * @param {?Function} error-first callback containing the validation errors if any.
+   */
+  validate(/* [key], callback */) {
+    const fallback = arguments.length <= 1 && typeof arguments[0] === 'function' ? arguments[0] : undefined;
+    const key = arguments.length <= 1 && typeof arguments[0] === 'function' ? undefined : arguments[0];
+    const callback = arguments.length <= 2 && typeof arguments[1] === 'function' ? arguments[1] : fallback;
+
+    const data = this.state;
```
gideonthomas: this is different from the actual mixin because in the form component, we store the data in the state (which was how it was for the previous version of the mixin) vs. as a return value of a function that the new version of the mixin requires.

templates/lib/validatorset.js
```diff
@@ -1,25 +1,56 @@
...
+      type: 'string',
+      pattern: regex.email
+    }],
+    label: 'Username or Email',
+    invalidMessage: 'Invalid username or email'
+  },
+  password: {
+    type: 'string',
+    label: 'Password',
+    pattern: /^\S{8,128}$/,
```
gideonthomas: I kept it as is because there seems to be two levels of validation.

templates/lib/validatorset.js
```diff
@@ -1,25 +1,56 @@
...
+      properties: {}
+    };
     fieldValues.forEach(function(entry) {
       var isDisabled = entry[Object.keys(entry)].disabled;
       Object.keys(entry).forEach(function(name) {
         if (!isDisabled && fieldValidators[name]) {
-          validators[name] = fieldValidators[name];
+          validators.properties[name] = fieldValidators[name];
+
+          if (name === 'feedback') {
```
gideonthomas: this is to maintain the previous functionality where only `feedback` was a required field.

templates/lib/validatorset.js
```diff
@@ -1,25 +1,56 @@
...
 var fieldValidators = {
-  username:       Joi.string().regex(regex.username).label('Username'),
-  uid:            Joi.alternatives().try(
-                    Joi.string().regex(regex.username).label('Username'),
-                    Joi.string().regex(regex.email).label('Email')
-                  ),
-  password:       Joi.string().regex(/^\S{8,128}$/).label('Password'),
-  email:          Joi.string().regex(regex.email).label('Email'),
-  feedback:       Joi.boolean().required().label('Feedback')
+  username: {
```
gideonthomas: this is json-schema

templates/lib/validatorset.js
```diff
@@ -1,25 +1,56 @@
...
+  feedback: {
+    type: 'boolean',
+    label: 'Feedback'
+  }
 };
 
 module.exports = {
   getValidatorSet: function (fieldValues) {
-    var validators = {};
+    var validators = {
```
gideonthomas: json-schema expects this format

test/mock-login/server.js
```diff
@@ -2,8 +2,8 @@ var Boom = require('boom');
 var Hapi = require('hapi');
 
 module.exports = function() {
-  var server = new Hapi.Server({ debug: false });
-  server.connection({
+  var server = new Hapi.Server({
```
gideonthomas: `server.connection` is not a thing anymore so we pass the host, port into the constructor.

test/mock-login/server.js
```diff
@@ -12,10 +12,10 @@ module.exports = function() {
     {
       method: 'POST',
       path: '/api/v2/user/verify-password',
-      handler: function(request, reply) {
+      handler(request, h) {
```
gideonthomas: all the functions now take `h` as the second parameter instead of `reply`.

test/mock-login/server.js
```diff
@@ -26,65 +26,65 @@ module.exports = function() {
         }
 
         if ( payload.uid === 'invalidResponse' ) {
-          return reply('not json');
+          return 'not json';
         }
 
-        reply(Boom.unauthorized('Invalid username/email or password'));
+        throw Boom.unauthorized('Invalid username/email or password');
```
gideonthomas: all errors inside the handlers should be thrown according to Hapi docs.

test/web.js
```diff
@@ -10,710 +10,801 @@ var testCreds = require("./testCredentials");
...
-  var s = server({
-    loginAPI: "http://localhost:3232",
-    cookieSecret: "test",
-    oauth_clients: testCreds.clients,
-    authCodes: testCreds.authCodes,
-    accessTokens: testCreds.accessTokens,
-    enableCSRF: false
-  });
+  var s, s2, ls;
+  const cookieSecret = "test".padEnd(32, "test");
```
gideonthomas: I changed the cookie secret here since the new hapi cookie secret requirements enforce a minimum of 32 characters.

test/web.js
```diff
@@ -10,710 +10,801 @@ var testCreds = require("./testCredentials");
...
 
-  var s2 = server({
-    loginAPI: "http://localhost:3232",
-    cookieSecret: "test",
-    oauth_clients: testCreds.clients,
-    authCodes: testCreds.authCodes,
-    accessTokens: testCreds.accessTokens,
-    enableCSRF: true
-  });
+  lab.before(async () => {
```
gideonthomas: I moved this into `lab.before` since the server module is now async.

test/web.js
```diff
@@ -10,710 +10,801 @@ var testCreds = require("./testCredentials");
...
 
-  var ls = loginServer({
-    loginAPI: "http://localhost:3232"
+    ls = loginServer({
+      loginAPI: "http://localhost:3232"
+    });
   });
 
-  lab.test("GET / Redirects to /signup", function(done) {
+  lab.test("GET / Redirects to /signup", async () => {
```
gideonthomas: you can skip reviewing most of the tests except for the ones where I comment on about a real change that I made.

test/web.js
```diff
@@ -10,710 +10,801 @@ var testCreds = require("./testCredentials");
...
+    Code.expect(response.headers["set-cookie"]).to.match(/crumb=/);
+
+    Code.expect(response.headers["strict-transport-security"]).to.equal("max-age=15768000");
+    Code.expect(response.headers["x-xss-protection"]).to.equal("1; mode=block");
+    Code.expect(response.headers["x-frame-options"]).to.equal("DENY");
+    Code.expect(response.headers["x-download-options"]).to.equal("noopen");
+    Code.expect(response.headers["x-content-type-options"]).to.equal("nosniff");
+
+    Code.expect(response.headers["content-security-policy"]).to.equal(
+      "base-uri 'self';" +
```
gideonthomas: I had to add this since it looks like `blankie` now adds this by default (doesn't seem to be harmful so I didn't remove it)

test/web.js
```diff
@@ -10,710 +10,801 @@ var testCreds = require("./testCredentials");
...
+      "connect-src 'self';default-src 'none';font-src 'self' https://fonts.gstatic.com;" +
+      "img-src 'self' data: https://www.google-analytics.com http://www.google-analytics.com;" +
+      "script-src 'self' 'unsafe-eval' https://www.google-analytics.com http://www.google-analytics.com;" +
+      "style-src 'self' https://fonts.googleapis.com"
+    );
+  });
+
+  lab.test("POST Create User", async () => {
+    try {
```
gideonthomas: this pattern for launching the login server happens in all the tests that need it.

web/index.js
```diff
@@ -18,10 +18,11 @@ var options = {
   redisUrl: process.env.REDIS_URL
 };
 
-var server = require('./server')(options);
+const start = async () => {
```
gideonthomas: made it a function since `await` can only be used inside async functions and not inside the global scope

web/server.js
```diff
@@ -17,14 +16,14 @@ var passTest = new PassTest({
...
-module.exports = function(options) {
+module.exports = async function server(options) {
   var serverConfig = {
     debug: options.debug,
-    connections: {
-      routes: {
-        security: true
-      }
-    }
+    routes: {
```
gideonthomas: changes in the API so now we have to pass the host/port directly to the constructor

web/server.js
```diff
@@ -39,31 +38,24 @@ module.exports = function(options) {
...
-  server.connection({
-    host: options.host,
-    port: options.port
-  });
-
   if ( options.logging ) {
-    server.register({
-      register: require('hapi-bunyan'),
+    await server.register({
+      plugin: require('hapi-bunyan'),
```
gideonthomas: the key for registering plugins is now called `plugin`

web/server.js
```diff
@@ -39,31 +38,24 @@ module.exports = function(options) {
...
       options: {
         logger: require('bunyan').createLogger({
           name: 'id-webmaker-org',
           level: options.logLevel
         })
       }
-    }, function(err) {
```
gideonthomas: we don't try catch any of these anywhere since the previous logic basically just threw the error which is what will happen now as well.

web/server.js
```diff
@@ -87,28 +79,26 @@ module.exports = function(options) {
         fontSrc: [
           '\'self\'',
           'https://fonts.gstatic.com'
-        ]
+        ],
+        generateNonces: false
```
gideonthomas: apparently blankie now auto-generates a few nonces that are injected into the csp. No idea what those nonces are used for but I disabled it.

web/server.js
```diff
@@ -129,27 +119,25 @@ module.exports = function(options) {
...
-    register: require('crumb'),
+  await server.register({
+    plugin: require('crumb'),
     options: {
       restful: true,
       skip: !options.enableCSRF ? skipCSRF : undefined,
       cookieOptions: {
-        isSecure: options.secureCookies
+        isSecure: options.secureCookies,
+        isSameSite: false,
```
gideonthomas: disables the `SameSite` header (which used to be the default behaviour before) which apparently if enabled, does not allow access to the header on the client (Login might be using it?)

web/server.js
```diff
@@ -129,27 +119,25 @@ module.exports = function(options) {
...
+        isHttpOnly: false
```
gideonthomas: also removes the HttpOnly header (default behaviour before)

web/server.js
```diff
@@ -183,7 +171,7 @@ module.exports = function(options) {
     {
       method: 'GET',
       path: '/login/oauth/authorize',
-      config: {
+      options: {
```
gideonthomas: API change to the key

web/server.js
```diff
@@ -212,71 +200,68 @@ module.exports = function(options) {
               redirect.query.state = request.query.state;
               redirect.query.scopes = request.query.scopes;
 
-              reply().takeover().redirect(url.format(redirect));
+              return h.redirect(url.format(redirect)).takeover();
```
gideonthomas: redirects and takeovers are now done through the `h` toolkit

web/server.js
```diff
@@ -325,8 +310,9 @@ module.exports = function(options) {
               otherwise: Joi.forbidden()
             })
           },
-          failAction: function(request, reply, source, error) {
-            reply(Boom.badRequest('invalid ' + source + ': ' + error.data.details[0].path));
+          failAction(request, h, error) {
+            const { source , keys: [ key ] } = error.output.payload.validation;
```
gideonthomas: the way we access validation failures has changed so we access `source` through a different object.

web/server.js
```diff
@@ -336,388 +322,345 @@ module.exports = function(options) {
...
+              return json.user;
             }
           }
         ]
       },
-      handler: function(request, reply) {
-        request.auth.session.set(request.pre.user);
-        reply({ status: 'Logged In' });
+      handler(request) {
+        request.cookieAuth.set(request.pre.user);
```
gideonthomas: API change: the session cookie is now accessed through `request.cookieAuth`

web/server.js
```diff
@@ -336,388 +322,345 @@ module.exports = function(options) {
...
+      options: {
         auth: false,
         plugins: {
           crumb: false
         },
         cors: true,
         validate: {
           payload: {
-            username: Joi.string().regex(/^[a-zA-Z0-9\-]{1,20}$/).required(),
+            username: Joi.string().regex(/^[a-zA-Z0-9\-]{1,20}$/).required(), // eslint-disable-line no-useless-escape
```
gideonthomas: i'm not sure why eslint complains about escaping `-` inside the character class (I'm pretty sure it's necessary) but I didn't want to change it and possibly break something.

lib/oauth-db.js
```diff
@@ -1,18 +1,22 @@
 var Boom = require('boom');
 var crypto = require('crypto');
 var Hoek = require('hoek');
-var pg = require('pg');
+const { Pool } = require('pg');
```
gideonthomas: updated the pg api

web/server.js
```diff
@@ -39,31 +38,42 @@ module.exports = function(options) {
...
 
-  server.connection({
-    host: options.host,
-    port: options.port
-  });
-
   if ( options.logging ) {
-    server.register({
-      register: require('hapi-bunyan'),
+    await server.register({
```
gideonthomas: changed from `hapibunyan` to `good-bunyan`. The logging looks kinda gross but it works.

web/server.js
```diff
@@ -39,31 +38,42 @@ module.exports = function(options) {
...
+              log: '*',
+              error: '*',
+              request: '*'
+            }, {
+              logger: require('bunyan').createLogger({
+                name: 'id-webmaker-org',
+                level: 'error'
+              }),
+              levels: {
+                response: 'info',
```
gideonthomas: feel free to suggest other mappings.
The mapping follows the format:
<good event>: <bunyan level level>

Note the the `log` good event is error because according to their docs, they also use this event for system level errors.

web/server.js
cadecairos: Tip: `npm install -g bunyan` then `npm start | bunyan` will format the output a bit better for debugging.

web/server.js
cadecairos: yeah, these broke donate too when I updated blankie. The nonces are supposed to let you allow inline scripts via CSP, I think.

web/server.js
cadecairos: Pretty sure we don't want the CSRF crumb cookie accessible over Javascript?

PR #448, opened 2018-07-18
Recaptcha for signups

cadecairos:


cadecairos:
@gideonthomas This is ready now. Testing is a bit of a pain because the webpack bundle task need environment variables set manually to add things to the app script. Here's how I did it:

`RECAPTCHA_SITE_KEY={a site key} RECAPTCHA_DISABLED=false ./node_modules/.bin/webpack --config webpack.config.js --progress --profile --colors`

Then run the server with `npm run server` (`npm start` will destroy the bundle you just built)

Also note that this is using the beta reCAPTCHA v3 api, which allows for a promise based method to get the recaptcha token to verify server-side. This was required because v2 required a callback to be specified on setup, and this caused brain-numbing scope issues getting the token in the form submit callback.

We'll need to let this run for a bit, determine what kind of scores we're seeing for the "signup" action, and set a score threshold, as well as verify the "success" param. (see the v3 docs [here](https://developers.google.com/recaptcha/docs/v3))

PR #459, opened 2019-07-24
Remove account creation link in header

gideonthomas:
Mainly for thimble but might as well do this entirely since no other service uses this and I didn't want to deal with figuring out how to make this configurable in the id code.


patjouk:
I think I'm following what's happening here. Are you also removing the `create an account` link from thimble?
I understand that you don't want to revisit this code, but what we saw with the learning.mozilla.org shutdown is that people will find way to register if they really want to (cf https://github.com/mozilla/learning.mozilla.org/pull/2561) :D The signup link might be available somewhere else and it would be better to remove that page or to replace the form by a "we're shutting down" message.

gideonthomas:
@patjouk I see where you're coming from (sorry I should have posted the same context here that I did in the Thimble pr).

The goal of this simply to deter people from creating new accounts, not to prevent them from doing it. That's the strategy I used with the Thimble PR as well. If we want to prevent them completely, it would require a lot of work because Thimble is also localized, and we don't have any l10n resourcing anymore.

If people find sophisticated ways to create accounts (like somehow generating the client code, etc. and crafting the URL to the signup page), that's fine as long as the volume of creations is low.

patjouk:
Works for me then :) Thanks for the explanation!

PR #460, opened 2019-07-30
Peg scooter version to patch releases

gideonthomas:
I noticed that staging crashes with an error with dependencies saying blankie depends on scooter to run.

Apparently both blankie and scooter added a breaking change in their minor release 🙄 where you need to refer to each package as @hapi/packagename.

Instead of updating those and figuring out what will break, I'm pegging both of their versions to when they worked with the original name.

### DIFF
--- a/test/web.js
+++ b/test/web.js
@@ -84,7 +84,7 @@ lab.experiment("OAuth", function() {
       "img-src 'self' data: https://www.google-analytics.com http://www.google-analytics.com;" +
       "script-src 'self' 'unsafe-eval' https://www.google-analytics.com http://www.google-analytics.com " +
       "https://www.google.com/recaptcha/api.js https://www.gstatic.com/recaptcha/api2/;" + 
-      "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;worker-src 'self'"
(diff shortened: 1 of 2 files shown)


test/web.js
```diff
@@ -84,7 +84,7 @@ lab.experiment("OAuth", function() {
       "img-src 'self' data: https://www.google-analytics.com http://www.google-analytics.com;" +
       "script-src 'self' 'unsafe-eval' https://www.google-analytics.com http://www.google-analytics.com " +
       "https://www.google.com/recaptcha/api.js https://www.gstatic.com/recaptcha/api2/;" + 
-      "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;worker-src 'self'"
+      "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com"
```
gideonthomas: I added this change when we updated blankie here: https://github.com/mozilla/id.webmaker.org/pull/447#discussion_r161321432
because it was another breaking change they added in a minor version. Since this pr lowers the version, this change had to be removed.

patjouk (approved):
:shipit:

